Issue #68, opened 2018-08-22
Dataset filelisting not updated when switching datasets


stephan271:
## Issue Name
FilelistingUpdate

### Summary
When switching between Datasets and entering the detail view showing the Data files the file listing is not updated, but the previous filelisting is shown

### Steps to Reproduce
See above

### Expected Behaviour
Datafile listing is always the correct one associated with the dataset

### Extra Details
A reload fixes the problem. The right data seems to be fetched in the background but just not being displayed

PR #71, opened 2018-08-23
Issue #68 fixed datablocks state management bug

lukegorman:
## Description

very small change to dataset reducers. Relevant action not referenced by reducer.

[x] Passing? (Merge will not be approved unless this is checked) 
[]  Docs updated?


stephan271:
Looks good to me, behaviour is now as I would I expect it

Issue #77, opened 2018-09-04
Error snackbar fails to clear after failed login


lukegorman:
Error snackbar fails to clear after failed login

Also the message seems to be empty.

lukegorman:
add timeout functionality?

garethcmurphy:
Fixed in PR #139

PR #139, opened 2018-11-19
Issue #77 force definition of duration param

lukegorman:
force definition of duration param

[x] Passing? (Merge will not be approved unless this is checked) 
[]  Docs updated?


lukegorman:
for some reason the default value for duration was being lost and was undefined when calling `snackBar.open ` in app.component.

PR #167, opened 2018-12-06
Dataset filters

JonasRosenqvist:
### Changes to dataset/filters

- Preventing form submit when pressing enter in the 'Text search'.
- For' Location', 'Group', 'Type' and 'Keywords' filters:
- - Show all suggestions when first selecting the (empty) field
- - When typing text, show only suggestions that match (contains) the typed text.
- - When pressing enter, automatically add the first (top) suggestion.
- - When adding a suggestion, clear the corresponding text field
- - After having added a suggestion, exclude this from future suggestions.

### DIFF
--- a/src/app/datasets/datasets-filter/datasets-filter.component.html
+++ b/src/app/datasets/datasets-filter/datasets-filter.component.html
@@ -22,10 +22,10 @@
           }} <mat-icon matChipRemove>cancel</mat-icon>
         </mat-chip>
       </mat-chip-list>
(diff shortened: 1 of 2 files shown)


hannes-petri-maxiv-lu-se:
Looking very good! Can I suggest:

1. Slightly more descriptive names for the subject, e.g. `locationInput$` instead of `locationInpuKeyUp$`
2. Factoring out the filter map? As far as I can see it's identical in all cases (except for variable names)

JonasRosenqvist:
I've broken out the duplicate code into a new function (createSuggestionObserver) and renamed the variables.

Issue #184, opened 2018-12-20
dont GET dataset attachments if items per page > 50


lukegorman:
## Issue Name
dont GET dataset attachments
### Summary

Users reporting problems when setting items per page = 1000.

need a more efficient method before we can roll this out unilaterally.

PR #185, opened 2018-12-20
Issue #184 dont get images if...

lukegorman:
## Description
as Issue #184 

[] Passing? (Merge will not be approved unless this is checked) 
[]  Docs updated?


lukegorman:
temporary change

Issue #208, opened 2019-01-15
Change detection problems in dataset details


lukegorman:
## Issue Name
Change detection
### Summary
On dataset click a number of components are loaded. Binding needs to be reviewed to ensure change detection is not throwing errors.

DatafilesComponent.html:2 

ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'ngPlural: 0'. Current value: 'ngPlural: 7272'.

PR #214, opened 2019-01-18
fixed console error about expression chaning after being checked

JonasRosenqvist:
Issue #208 
also fixed console error relating to jwt.


lukegorman:
lovely job

PR #223, opened 2019-01-23
new pipe for stripping non-numeric characters from a string

JonasRosenqvist:
So I solved it with a new pipe, that strips non-numeric characters from a string. I'm not sure if that's an overkill for such a simple task, but I trying to follow the general conventions, which appears to be using pipes for these types of transformations.

The name of the pipe might be a bit misleading as well; I was trying to with something short.


lukegorman:
probably would have just trimmed it but this is good too.

PR #225, opened 2019-01-29
defined the proposal id pipe to strip given prefix instead of all non…

JonasRosenqvist:
It turned out ESS were using non-numeric characters in their proposal id, so instead we just strip the prefix "proposal-" (if present).


lukegorman:
ERROR TypeError: Cannot read property 'replace' of undefined
    at StripProposalPrefixPipe.push../src/app/shared/pipes/stripProposalPrefix.pipe.ts.StripProposalPrefixPipe.transform (stripProposalPrefix.pipe.ts:10)
    at checkAndUpdatePureExpressionInline (core.js:22697)
    at checkAndUpdateNodeInline (core.js:23269)
    at checkAndUpdateNode (core.js:23227)
    at debugCheckAndUpdateNode (core.js:23861)
    at debugCheckRenderNodeFn (core.js:23847)
    at Object.eval [as updateRenderer] (DatasetTableComponent.html:242)
    at Object.debugUpdateRenderer [as updateRenderer] (core.js:23839)
    at checkAndUpdateView (core.js:23214)
    at callViewAction (core.js:23450)

lukegorman:
check for null?

JonasRosenqvist:
added!

Den tis 29 jan. 2019 kl 13:31 skrev Luke Gorman <notifications@github.com>:

Issue #217, opened 2019-01-23
Delete Attachments individually


stephan271:
## Issue Name
DeleteAttachmentsIndividually

### Steps to Reproduce
Add two attachments, then delete the first. 

### Current Behaviour
Both attachments are deleted

### Expected Behaviour
Only the first should be deleted

PR #269, opened 2019-02-28
Feature/allow users to add attachments

stephan271:
## Description
Allow users to add attachments

## Motivation
Currently only the ingestor could add datasets

## Fixes:
#217 

## Changes:

Uses the indirect API via the Dataset model rather than the direct model to DatasetAttachment. Therefore the direct access can now be "closed" for all users in catamel to avoid an open access to attachments


stephan271:
Could one of you have a look at the test failure messages > My tests here locally succeed

lukegorman:
im happy that this will build after merge

PR #398, opened 2019-07-16
Share data with other users

lukegorman:
## Description

add the other user to a group, which you add to the accessGroups list. Then catanie would need to make a call to update the dataset by adding this group to the dataset's accessGroups and another can to create the group and its members. When the other user re-login he would see the dataset. One would need to discuss were the groups can be managed. This could be either in an IM System like AD or we would create a new group model inside loopback, which would manage these kind of ad hoc groups (i.e re-vive the accessGroups model or somethng similar). I guess this needs some more discussions.

[] Included for each change/fix?
[] Passing? (Merge will not be approved unless this is checked) 
[] Docs updated?
[] Toggle added for new features?


lukegorman:
WIP: just created MR for convenience

lukegorman:
this MR was in error...

PR #591, opened 2020-01-07
add sorting to jobs, fix sorting direction (should be space separated)

lukegorman:
## Description
add sorting to jobs, fix sorting direction (should be space separated)

issue #590 

- [ x] Passing? (Merge will not be approved unless this is checked) 
- [ ] Docs updated?


stephan271:
It would be good to have the sorting option for all columns. My npm run test run results in 6 errors

Issue #593, opened 2020-01-09
datafiles table not showing records correctly


lukegorman:
### Summary

for some DSs the path column is not populated and for others, no records are shown even though the API call has clearly returned. 

Furthermore conditionals in datafiles.html are unreadable and need to be rationalized. There is no clear way to disable file download functionality.

PR #594, opened 2020-01-09
Hotfix/data files

lukegorman:
## Description

data files table now works without file download functionality
fixes issue #593


henrikjohansson712:
Looks good! Just change multipleDownloadEnabled back to true in environment.ts

PR #603, opened 2020-01-22
Added encodeURIComponent to logbook effects

### LINKED ISSUE #599: properly toggle logbook functionality
dataset detail throws errors because logbook fields are null.

this shouldnt happen when logbook is toggled off

henrikjohansson712:
## Description

Wrapped proposalId in encodeURIComponent for logbook fetches.

## Motivation

See issue #599 

## Fixes:

* Fixes #599 

## Changes:

* Wrapped proposalId in encodeURIComponent for logbook fetches.

- [ ] Docs updated?

### DIFF
--- a/src/app/state-management/effects/logbooks.effects.ts
+++ b/src/app/state-management/effects/logbooks.effects.ts
@@ -35,13 +35,15 @@ export class LogbookEffects {
       ofType(fromActions.fetchLogbookAction),
       withLatestFrom(this.filters$),
       mergeMap(([{ name }, filters]) =>
-        this.logbookApi.filter(name, rison.encode_object(filters)).pipe(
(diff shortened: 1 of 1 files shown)


lukegorman:
Question: could we use rison to encode the name as well?

henrikjohansson712:
The logbook name is the proposal id of the connected proposal. We're using encodeURIComponent for that everywhere else, so I can't really see the reason for using a different approach here.

lukegorman:
> The logbook name is the proposal id of the connected proposal. We're using encodeURIComponent for that everywhere else, so I can't really see the reason for using a different approach here.

because in the same line we are using rison
filter(encodeURIComponent(name), rison.encode_object(filters))

PR #672, opened 2020-06-30
Hotfix/file dl

lukegorman:
## Description

Restore file download functionality previously removed as part of bug fix

- [ ] Docs updated?

### DIFF
--- a/src/app/datasets/datafiles/datafiles.component.ts
+++ b/src/app/datasets/datafiles/datafiles.component.ts
@@ -27,7 +27,7 @@ import { UserApi } from "shared/sdk";
   templateUrl: "./datafiles.component.html",
   styleUrls: ["./datafiles.component.scss"],
 })
-export class DatafilesComponent implements OnInit, OnDestroy, OnChanges {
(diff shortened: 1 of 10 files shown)


src/app/datasets/datafiles/datafiles.component.ts
```diff
@@ -56,36 +60,92 @@ export class DatafilesComponent implements OnInit, OnDestroy, OnChanges {
...
     private cdRef: ChangeDetectorRef,
+    private userApi: UserApi,
+
     @Inject(APP_CONFIG) public appConfig: AppConfig
   ) {}
 
-  getDatafiles(datablocks: Array<OrigDatablock>) {}
-
-   hasTooLargeFiles(files: any[]) {
+  hasTooLargeFiles(files: any[]) {
```
henrikjohansson712, line 148: * This method and `onPageChange` below should also be moved above the `constructor`
* `onRowClick` is unused, should be removed

lukegorman:
> Just one last fix, then it should be good to go. I like the solution for the pipes!

sorry, i missed that. ill make that change in develop

PR #686, opened 2020-09-23
Also datablock file infos needs be removed from JSON display

stephan271:
## Description

Avoid display of large file arrays in JSON metadata display

## Motivation

Stabilize browser by avoiding display of large of large filearrays i

## Changes:

* in additon to origdatablocks the datablocks information is removed as well


henrikjohansson712 (approved):
Looks good!

Issue #475, opened 2019-09-04
search samples


toqduj:
Searching for samples based on its metadata or sample characteristics would be nice.

henrikjohansson712:
With these features it's now possible to filter samples on sample characteristics in the same way as you can filter datasets on scientific metadata. It does, however, require that the samples characteristics are stored in the following format:
```json
{
    "sampleCharacteristics": {
        "mass": {
            "type": "measurement",
            "value": 5,
            "unit": "grams",
            "valueSI": 0.005,
            "unitSI": "kg"
        },
        "volume": {
            "type": "measurement",
             "value": 1,
             "unit": "cm^3",
             "valueSI": 0.0000010000000000000002,
             "unitSI": "m^3"
        }
    }
}
```
The "type" field is necessary for now, but we're working in removing it. The SI value and unit will be calculated and added automatically by the backend upon saving with the following request:

`PATCH /Samples/{id}`
```json
{
    "sampleCharacteristics": {
        "mass": {
            "type": "measurement",
            "value": 5,
            "unit": "grams"
        },
        "volume": {
            "type": "measurement",
             "value": 1,
             "unit": "cm^3"
        }
    }
}
```

PR #703, opened 2020-11-20
Feature/search sample characteristics

henrikjohansson712:
## Description

Add filtering on sample characteristics to sample table.

Requires [catamel #349](https://github.com/SciCatProject/catamel/pull/349)

## Motivation

#475 

## Changes:

* Create new shared Search Parameters Dialog module
* Add button for characteristics filtering to sample dashboard
* Add chip list for applied characteristics filters to sample dashboard
* Update sample sdk with new metadataKeys endpoint
* Rename "Add Sample" button to "Create Sample", for consistency with dataset equivalent and to avoid confusion with new "Add Characteristic" filter
* Create GenericFilter interface containing filters that are common among several models

- [ ] Docs updated?
- [x] Requires update of catamel API?

### DIFF
--- a/src/app/policies/policies-dashboard/policies-dashboard.component.spec.ts
+++ b/src/app/policies/policies-dashboard/policies-dashboard.component.spec.ts
@@ -29,7 +29,7 @@ import {
 } from "state-management/actions/policies.actions";
 
 import { Router } from "@angular/router";
-import { PolicyFilters } from "state-management/models";
+import { GenericFilters } from "state-management/models";
 import * as rison from "rison";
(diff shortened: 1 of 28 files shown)


stephan271:
Is there a flag in environments related to this feature , which I need to switch on for tests ? Do we need to care for the "coverage" CI/CD complaints ?

henrikjohansson712:
Added the same flag as for the dataset version of the search: `scienceSearchEnabled`. The coverage always fluctuates a bit, but I realised that I forgot the unit tests for the new module, so I added those.

PR #725, opened 2021-02-22
Support to filter by date range

stephan271:
## Description

Support to filter by date range

## Motivation

Allow column definitions for date values in shared-table  to let users select the date range via calendar widgets


henrikjohansson712 (changes requested):
Fix DeepScan issue

stephan271 (commented):
Deepscan issues have been resolved

henrikjohansson712 (approved):

PR #753, opened 2021-04-01
Add OpenID connect Functionality

dylanmcreynolds:
## Description
Add ability authenticate through OIDC.

## Motivation
Use Case: my facility does not have users stored in LDAP/AD. We have a user system that stores information in a custom database, and allows users to authenticate through ORCID. So, to prevent users from having to create new accounts in Scicat, we would want ORCID authentication. ORCID exposes OIDC. 

To date, catanie can authenticate with two catemel/loopback/passport providers: local and ldap. Catanie communicates with catamel through loopback SDK methods. A user enters username and password in catanie, this gets sent to loopback, which responds with and access_token and user_id.

Passport has several other authentication strategies available, including OIDC. OIDC authentication differs from LDAP in several ways. One fundamental different is the the resource app (catamel/catanie) does not ever deal with the users authentication information. Instead, it redirects the user to the OIDC provider's website. On successful authentication, the OIDC provider redirects the user's browser back to the service provider's site. This means that catanie must interact in a slightly different way than it does for existing authentication.

## Changes:
* login-component.html - added a new button currently called "OIDC Login" that starts the authentication process
* login-component.ts 
   *  inject `Document` into the constructor, used to redirect the user
   * add method `redirectOIDC()` to redirect user to OIDC login
   * changed `ngOnInit()` adding a subscription to listen for  the part  of the url path that tells us an access_token is available
* user.actions.ts - adds `loginOIDCAction` 
* user.effects.tx - add new `oidcFetchUser$` to observe the `loginOIDCAction` and construct a new `SDKToken` object and complete the login.
Here is a high level view of how the flow works:
image

## What's not there yet?
I am submitting this incomplete to get people's eyes on it and tell me if this is sensible. Things that need to be done for sure:
- [ ] Handle login failure
- [ ] Get accessGroup information from the UserIdnetity collection
- [ ] Make this configurable...I can imagine some sites want only OIDC, only LDAP and maybe a combination of the two? Right now, both options are placed on the login form.

- [ ] Docs updated?

### DIFF
--- a/src/app/state-management/actions/user.actions.ts
+++ b/src/app/state-management/actions/user.actions.ts
@@ -2,6 +2,12 @@ import { createAction, props } from "@ngrx/store";
 import { User, AccessToken, UserIdentity, UserSetting } from "shared/sdk";
 import { Message, Settings } from "state-management/models";
 
+
+export const loginOIDCAction = createAction(
+  "[User] OIDC Login",
+  props <{accessToken: string; userId: string}>()
+)
+
 export const loginAction = createAction(
   "[User] Login",
   props<{ form: { username: string; password: string; rememberMe: boolean } }>()
--- a/src/app/state-management/effects/user.effects.ts
+++ b/src/app/state-management/effects/user.effects.ts
@@ -82,6 +82,28 @@ export class UserEffects {
     )
   );
 
+  oidcFetchUser$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(fromActions.loginOIDCAction),
+      switchMap(({ accessToken, userId }) => {
+        const token = new SDKToken({
+          id: accessToken,
+          userId: userId,
+        });
+        this.loopBackAuth.setToken(token);
+        return this.userApi.findById(userId).pipe(
+          switchMap((user: User) => [
+            fromActions.fetchUserCompleteAction(),
+            fromActions.loginCompleteAction({
+              user,
+              accountType: "external",
+            }),
+          ]),
+          catchError(() => of(fromActions.fetchUserFailedAction()))
+        );
+      })
+    )
+  );
   fetchUser$ = createEffect(() =>
     this.actions$.pipe(
       ofType(fromActions.fetchUserAction),
--- a/src/app/users/login/login.component.html
+++ b/src/app/users/login/login.component.html
@@ -3,11 +3,25 @@
     <div fxLayout="row" class="facility-logo">
       <div fxFlex="auto"></div>
       <div fxFlex="200px">
-        <img src="assets/images/site-logo.png" />
+        <img src="assets/images/esslogo.png" />
       </div>
       <div fxFlex="auto"></div>
     </div>
     <mat-card-content>
+
+        <button
+        mat-raised-button
+        class="login-button"
+        type="submit"
+        color="primary"
+        [ngClass]="{ loading: loading$ | async }"
+        (click) = "redirectOIDC()"
+      >
+       OIDC Log in
+      </button>
+    
+      
+      <!-- <a href="http://localhost:3000/auth/oidc"><button class="btn btn-success pull-right" >OIDC login</button></a> -->
       <form [formGroup]="loginForm" (ngSubmit)="onLogin()">
         <mat-form-field hintLabel="{{appConfig.facility}} account username">
           <mat-label> <b> Username</b> </mat-label>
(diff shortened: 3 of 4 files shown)


stephan271:
@dylanmcreynolds could you have  a look why the tests are failing ?

dylanmcreynolds:
@stephan271 Looks like two things. 

One, the E2E tests are having an issue logging on to dockerhub. I'm not familiar with this workflow...if it's coming from my fork do I need a dockerhub account so that I can place secrets somewhere?

Second, coveralls is rightly complaining that the oidc code in angular isn't tested. I'll address that...but I still need to update this code to make the appearance of various login methods in the login.component configurable.

dylanmcreynolds:
I added functionality to configure:

- whether to display the login form (based no environment setting `loginFormEnabled`
- a list of OIDC providers that can be configured by the environment setting oAuth2Endpoints

dylanmcreynolds:
I have  added configuration to the [login form](https://github.com/dylanmcreynolds/catanie/blob/d775afcf3114df6a97cd4df801b3c5d78ae3574b/src/app/users/login/login.component.html#L11) and I would like to add tests to src/app/users/login/login.component.spec.ts to test that these configurations from environment make the appropriate UI change: 
1. If loginFormEnabled is false, the form is not displayed
2. If an OAuth2Provider is configured, their button dsplayed and if clicked, performs the onLoginOIDC action

This involves a rearrangement of the test file that's a little beyond me at present. The settings can be mocked as they are [here](https://github.com/dylanmcreynolds/catanie/blob/d775afcf3114df6a97cd4df801b3c5d78ae3574b/src/app/users/login/login.component.spec.ts#L60), but this is done in a global describe/beforeEach. Should I have separate top level describes for these different cases @henrikjohansson712 ?

henrikjohansson712:
@dylanmcreynolds Something like this might be useful: https://github.com/angular/quickstart/issues/320#issuecomment-354471054

As I understand it, you can set up the basic config in the global scope, then override with the necessary values in the individual `describe` scopes.

dylanmcreynolds:
Hmm....I've added 
```
 describe("remove username/password", () => {
    beforeEach(() => {

      TestBed.overrideProvider(APP_CONFIG, {
        useValue: {
          disabledDatasetColumns: [],
          archiveWorkflowEnabled: true,
          loginFormEnabled: false
        } 
      });
      TestBed.compileComponents();
      fixture = TestBed.createComponent(LoginComponent);
      component = fixture.componentInstance;
      fixture.detectChanges();
    });
    it("should not contain username password fields", ()=> {
      const compiled = fixture.debugElement.nativeElement;
      expect(compiled.querySelector("form")).toBeNull();
    });
  });
```

but this complains:
" Error: Cannot override provider when the test module has already been instantiated. Make sure you are not using `inject` before `overrideProvider`. "

henrikjohansson712:
@dylanmcreynolds Could you push the changes so I can have a look at the file?

dylanmcreynolds:
@henrikjohansson712 I just pushed a [change](https://github.com/SciCatProject/catanie/pull/753/commits/5d9790c7ea2b0f007dc439fd259fbe8e2ffbe8e6). It's a little ungraceful, but it seems to work. It moves createComponent later so that each test can set the values I for `appConfig.loginFormEnabled` that I need for the tests.

dylanmcreynolds:
Progress on this PR is slow, but continuing. At least two things still need to be addressed:

- [ ] How to handle a failure to login? What do we want to display? Do we want to treat errors communicating with AuthProvider differently auth provider denying access?
- [ ] Google and ORCID both have requirements that buttons pointing the user to authenticate with their system have branded look and feel. Need to add some mechanism in the configuration section to provide styling/html.

bolmsten (approved):

dylanmcreynolds:
Test coverage is getting close, but I'm having a hard time figuring out how to get lines of code in app-config-module.ts to be recognized as being covered in src/app/app-config.module.spec.ts

dylanmcreynolds:
One commit that I made which people will need to note:

This [commit](https://github.com/SciCatProject/catanie/commit/fc565153602adb556ad48f58bb9d8e870e5ec0f8) from a month ago had, I believe, the side effect that any loaded environment.*.ts file must contain the entire list of environment properties. Otherwise they, you see a typesrcipt error from app-config-module.ts when loading the environment into APP_DI_CONFIG.

So, I [added](https://github.com/SciCatProject/catanie/pull/753/commits/dd3ebf9afc5f36d8c1b80e8f340a08567eb53d11) my two new fields to all of the environement.*.ts files.

dylanmcreynolds:
I'm not positive about the protocol for SciCat projects. I am hesitant to merge my own PR.

nitrosx:
I'm ready to merge this PR.
Does anybody have any reason for hold off the merge?
Please let me know.

stephan271:
Please go on, no objections from my side

PR #755, opened 2021-04-06
Scientific metadata view

nguyenlinhlinh:
## Description
This alternative view shows the metadata in tree structure which can be used for both scientific metadata that has flat structure or multi-level structure.  
This PR is also included some changes required due to enabling some angularCompilerOptions.
## Motivation
We have facility using flat data structure and facilities using multi-level data structure so this is an attempt to make it possible to view metadata regardless the different data structures. 
## Features:
- Showing both flat and multi-level data structure
-  User can edit/delete/add new metadata
-  User can redo/undo changes (Not after it have been saved to the DB)
-  User cannot edit fields that has value of type array
-  User can search/filter fields by text
-  Leaving page guard (User is asked for saving unsaved changes when leaving page).
## Usage:
Enable this view by adding the below fields to environment file. 
```
 tableSciDataEnabled: true,
 metadataStructure: "tree",
```
Note: if `metadataStructure` is not provided the default view is used so you don't need to do anything if you want to use the view for flat data structure.

- [ ] Docs updated?

I know that it fails on deepscan but the first one is intended and other unused functions and variables is not written by me so I don't want to remove them.


henrikjohansson712 (changes requested):
Everything works and layout looks nice, but there are some deepscan issues that needs looking in to.

henrikjohansson712 (approved):

Issue #770, opened 2021-04-30
Wrong field casting when editing metadata


minottic:
### Summary
When adding or editing string field and inserting numeric value, on save, field is converted to date, with value 1970/1/1...
This happens during editing of nested scientific metadata and probably is a side effect of #755 
### Steps to Reproduce
![image](https://user-images.githubusercontent.com/50220438/116736712-8c898500-a9f0-11eb-95de-6238b99d3889.png)
![image](https://user-images.githubusercontent.com/50220438/116736767-9ad7a100-a9f0-11eb-9070-c8fa218e249d.png)
![image](https://user-images.githubusercontent.com/50220438/116736800-a6c36300-a9f0-11eb-973e-79c0a1b80291.png)
### Current Behaviour
String field is converted to date, with value 1970/1/1.
### Expected Behaviour
Keep string and cast numeric value to string 
### Extra Details

Here you should include details about the system (if it is unique) and possible information about a fix (feel free to link to code where relevant). Screenshots/GIFs are also fine here.

nguyenlinhlinh:
Hi @minottic, I tried to reproduce the bug but it doesn't show up. Could you please check again to see if you have the latest changes? otherwise maybe you can show it to me on the meeting tomorrow.

minottic:
Hi @nguyenlinhlinh , dunno, maybe it then relates to something else. I will have a better look, and let's touch on it tomorrow during the meeting if I keep seeing the same behavior. Thanks!

nguyenlinhlinh:
@minottic I am suspecting that this [function](https://github.com/SciCatProject/catanie/blob/e751564a773f2c2943591619109dbaea3dd839ad/src/app/shared/modules/scientific-metadata-tree/base-classes/tree-base.ts#L190) is the one causing the bug but I can't not reproduce it. Could you check if the code for this function in your deployment looks the same?

nguyenlinhlinh:
Ah I could reproduce the bug now with value "1000" shall try to fix that

PR #773, opened 2021-05-12
Luxon

nguyenlinhlinh:
## Description

The PR fixes issue #770, fixes a bug in date filter for dataset and switches momentjs to luxon since momentjs is not recommended to use anymore (See [here](https://momentjs.com/docs/))

## Motivation
Using JS Date object to identify date causing problem as showed on issue #770, therefore I applied a stricter rule on how to identify a string as date (See [here](https://github.com/SciCatProject/catanie/blob/b7b47fca6587eb5f4c572e8b0613fc71ca5f152d/src/app/shared/services/date-time.service.ts)). This implementation assumes that all dates that are passing around between backend and frontend are in ISO format so a string can only be identified as date if it is in ISO format. For user input a string is valid date if it is in the following format: 
`yyyy-MM-dd`, `yyyy-MM-dd HH:mm`,`yyyy-MM-dd HH:mm:ss`

## Fixes:
- Fixed a bug in date filter for dataset. When user filters datasets using date filter and later on clear the date filter by erasing the input field instead of using clear button the result is empty.
-  Fixed the formatting issue in #770 

## Changes:
- Switch from momentjs to luxon
- Make sat-date-picker using luxon DateTime

- [ ] Docs updated?


src/app/files/files-dashboard/files-dashboard.component.ts
```diff
@@ -12,15 +11,13 @@ import { Column } from "shared/modules/shared-table/shared-table.module";
...
-  end = moment();
-  start = moment().subtract(7, "days");
 
   columns: Column[] = [
     { id: "dataFileList.path", icon: "text_snippet", label: "Filename", canSort: true, matchMode: "contains", hideOrder: 1, },
     { id: "dataFileList.size", icon: "save", label: "Size", canSort: true, matchMode: "greaterThan", hideOrder: 2, },
     {
-      id: "dataFileList.time", icon: "access_time", label: "Created at", format: "date medium", canSort: true,
-      matchMode: "between", filterDefault: { start: this.start.format("YYYY-MM-DD"), end: this.end.format("YYYY-MM-DD") }, hideOrder: 3,
+      id: "dataFileList.createdAt", icon: "access_time", label: "Created at", format: "date medium", canSort: true,
```
henrikjohansson712, line 19: The change from `"dataFilelist.time"` to `"dataFileList.createdAt"` causes the *Created At* column to be empty. Looks like all the fields like `"createdAt"` and `"updatedAt"` are removed before the data gets loaded into the table, both on the items in `dataFileList` and in the `origDatablock` itself. @stephan271 do you know why?

stephan271: I think there is no createdAt field for the individual files. Why was it changed from time to createdAt ?

henrikjohansson712: There is on the ESS ones at least:

```json
  {
    "id": "000ae0b6-bddc-4b19-a86d-f5e1a6a161b9",
    "size": 0,
    "ownerGroup": "ess",
    "accessGroups": [
      "ess",
      "odin",
      "loki",
      "nmx"
    ],
    "createdBy": "ingestor",
    "datasetId": "20.500.12269/1073nicos_00001073.hdf",
    "dataFileList": [
      {
        "path": "nicos_00001073.hdf",
        "size": 0,
        "time": "2019-11-26T12:46:51.000Z",
        "chk": "34782",
        "uid": "101",
        "gid": "101",
        "perm": "755",
        "createdAt": "2021-05-18T08:46:50.700Z"
      }
    ],
    "createdAt": "2019-11-26T12:47:01.479Z",
    "updatedAt": "2019-11-26T12:47:01.480Z"
  }
```
But this is what's loaded into the table:

```json
{
    "dataFileList": {
        "path": "nicos_00001073.hdf",
        "size": 0,
        "time": "2019-11-26T12:46:51.000Z",
        "chk": "34782",
        "uid": "101",
        "gid": "101",
        "perm": "755"
    },
    "ownerGroup": "ess",
    "datasetId": "20.500.12269/1073nicos_00001073.hdf",
    "id": "000ae0b6-bddc-4b19-a86d-f5e1a6a161b9",
    "uniqueId": 1
}
```

henrikjohansson712: I think I know why now. The Datafile model extends the MongoQueryable model, which in turn extends the Ownable model. So the `createdAt` on the items in the `dataFileList` gets added automatically with the current datetime, so it shows up on queries but is not actually stored in the database. So this should be changed back to use the `time` field.

henrikjohansson712 (changes requested):
The change from `dataFileList.time` to `dataFileList.createdAt` should be reverted. Other than that everything seems to work correctly.

src/app/files/files-dashboard/files-dashboard.component.ts
nguyenlinhlinh: The reason I change to `createAt` was because we got empty Create At column since we don't have `time` field. I will revert it.

henrikjohansson712 (approved):

PR #776, opened 2021-05-17
styled shared table header

nitrosx:
## Description

Style shared table header consistently with dataset table

## Motivation

Shared table header was not styled as the main datasets table. As we would like to leverage the shared table component for all the tables, a restyle was needed. 

## Changes:

* Added class to the shared table html
* changed css style for shared table in component scss file
* added styles in main theme.scss

- [ ] Docs updated?

## Notes

I'm not sure that I place the css styles in the appropriate files. Please review them and let me know if I need to relocate them in different ones.


stephan271:
The curent styling makes the filter condition very hard to read (blue on blue) . Can this be changed ?
image

nitrosx:
> The curent styling makes the filter condition very hard to read (blue on blue) . Can this be changed ?
> image

Sure!!! Would white or light gray be ok?

stephan271:
Anything having enough contrast is fine for me, no strong preferences. Thanks!

stephan271:
I just compared to the old state:
image

I see that the Job title bar seems to have disappeared. Also, as an alternative approach: couldnt we just turn the gray areas in the above screenshot into blue ones, but avoid to have a full bar across the whole width ?

nitrosx:
@stephan271: as a test, I remove the title as there is the breadcrumbs that already indicates which page we are looking at. I can reinstate it, but it will take space away from the search bar.
Regarding making just the filter areas blue, I'm fine with it, but than we woul dneed to change our dataset page to match data.
Would a lighter background color across the whole header be more acceptable?
I could try to prepare couple of images examples, so we can decide before changing the css

stephan271:
I think preparing a few images would be great, then we can discuss it, may be via zoom ? Thanks for your effort .

nitrosx:
I updated the shared table header based on the feedback from all of us.
The header color will change according to the color schema selected in the local theme.

In a vanilla scicat installation, the shared table header will look like as follow:
shared_table_v4

No title in the table header. I would suggesto to add the title of the table in the page title in the top banner or just use the breadcrumb.

stephan271:
Concerning the title: may be we can make this configurable. There may be situation where the table (since it is a standalone component) is used in another environment, where it will be useful to see what is actually displayed, and have this iinformation in close neighborhood to the table. Think e.g. of a case, where you have two tables showing different quantities next to each other. So could we define a flag, which makes the title appear or disappear ?

nitrosx:
@stephan271 the use case that you mention makes perfect sense, and I think it is totally doable. Let me check the code and see how it can be implemented.

stephan271:
LGTM, Tests ran fine locally at PSI. Thanks for the effort. From my point of view this PR can be merged.

PR #786, opened 2021-06-01
Issue777 - editPublishedData default

### LINKED ISSUE #777: editPublishedData cannot be overridden to false
### Summary
When trying to create a new boolean setting in https://github.com/SciCatProject/catanie/blob/develop/src/app/app-config.module.ts, I followed the pattern that is used:

```
editPublishedData: environment["editPublishedData"] || true,
```
I believe that this is intended to say "if not present in environment, return a default of true"

However, if the environment at that key has a `false` value, the statement still returns `true` because  ` false || true` evaluates to `true`

This may not be a big issue, because `editPublishedData` is the only setting where the default is trying to be set to true and I don't see `editPublishedData` used anywhere. However, I think it would be fixable by changing that to:
```
editPublishedData: (environment["editPublishedData"] === undefined) ? true: environment["editPublishedData"] 
```
or removed, if it's not being used.

dylanmcreynolds:
## Description

Fixes #777  

When working on creating settings in app-config.module.ts in a feature PR, I found I copied the setting for `editPublishedData` and found that the value was always true.

The way these settings are written, it appears that they are intended to have a default. However, the way `editPublishedData` is written, the value can never be false, since `false || true == true`

Additionally, it appears that most projects are using the `lbBaseURL`, but it is not defiend in `AppConfig`. So, for consistency, I added this field.

## Motivation

Fix it so that the code that uses editPublishedData can set to false if wanted.

## Fixes:

- added `lbBaseURL` to `AppConfig`
- fixed config logic so that `editPublishedData` can be set to false if configured

I added tests to the place that `editPublishedData` is used, to test that it's affecting the UI. But as I type this, I'm guessing that my real addition (change to the construction of how that variable is set in `AppConfig` isn't tested. Not quite sure how such an angular test would be constructed.


dylanmcreynolds:
With apologies...this PR was based on a poor merge. It has been superseded by #787 .  Closing

PR #800, opened 2021-07-05
datepicker for scientific metadata

nguyenlinhlinh:
## Description

Input field of type datetime-local is supported differently on Chrome vs Firefox (on Chrome a datepicker shows up on the input field but not on Firefox) therefore I choose to use material datetime-picker instead.

## Changes:
- Use datetime-picker

- [ ] Docs updated?


henrikjohansson712 (approved):

minottic (approved):

minottic:
quick comment, I think tests were added and are passing, so maybe worth ticking the boxes

PR #802, opened 2021-07-08
Show values in custom column

nguyenlinhlinh:
## Description
There are custom columns that show values in requested unit. This works for flat metadata but not nested. This PR adds support for nested metadata.
Screenshot 2021-09-15 at 09 55 53

- [ ] Docs updated?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2880,6 +2880,11 @@
       "integrity": "sha512-YSBPTLTVm2e2OoQIDYx8HaeWJ5tTToLH67kXR7zYNGupXMEHa2++G8k+DczX2cFVgalypqtyZIcU19AFcmOpmg==",
       "dev": true
     },
+    "@types/lodash": {
+      "version": "4.14.171",
+      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.171.tgz",
+      "integrity": "sha512-7eQ2xYLLI/LsicL2nejW9Wyko3lcpN6O/z0ZLHrEQsg280zIdCv1t/0m6UtBjUHokCGBQ3gYTbHzDkZ1xOBwwg=="
+    },
     "@types/luxon": {
       "version": "1.27.1",
       "resolved": "https://registry.npmjs.org/@types/luxon/-/luxon-1.27.1.tgz",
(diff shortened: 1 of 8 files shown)


stephan271:
@nguyenlinhlinh what is the status for this merge request ? There seem to be some issues. Do you need any feedback from me or Henrik before being able to address the failing checks ?

nguyenlinhlinh:
@stephan271  I will take a look into this

nguyenlinhlinh:
Hi @stephan271, now this PR is redo for reviewing/merging again

stephan271:
Does not yet seem to work for our data: 
image

If you want we can have a zoom session later today, then I can show you how it looks on our side.

nguyenlinhlinh:
@stephan271  Could it be that your data is save as {v: value, u: unit} instead? Yes we can talk over zoom. I am free the whole day.

stephan271:
Yes it is in v,u format. Can this be supported as well ?

nguyenlinhlinh:
@stephan271 yes, I can add support for that. I think it also need some changes in the search function in the backend to search using v and u. However it would be nice if you can migrate to values and unit :) so we a standard.

stephan271:
I am afraid we currently have a mix of both. Changing to one standard would mean, to update thousands of datasets. So may be extending the "standard" to cover both cases is the more pragmatic way, especially since many of the raw data files (hdf5) also have the v/u syntax already, and those can not/will not be changed.

nguyenlinhlinh:
@stephan271  I see, then it is easier to change in the code instead. could you try again to see if it works.

stephan271:
Sorry, it does not yet work on my side. I get the following error in the browser console as soon as I select some of the quantities in the scientificMetadata section

> ERROR TypeError: dataset_r96.scientificMetadata is undefined
>     DatasetTableComponent_ng_template_47_ng_template_0_mat_cell_2_Template dataset-table.component.html:362
>     Angular 40
>     RxJS 5
>     Angular 9
> core.js:6479
>     Angular 14
>     RxJS 5
>     Angular 9
>

nguyenlinhlinh:
@stephan271 hmm, it is difficult for me to debug when I don't see how the dataset looks like. Maybe we can talk over zoom when you have time. I am suspecting that the dataset doesn't have **scientificMetadata** field.  **dataset_r96** is strange I wonder where it comes from.

nguyenlinhlinh:
Hi @stephan271, I have added the tooltip as you suggested. However, I don't know how to fix the failed unit test that complain about coverage. I ran the test locally and it worked for me.

nguyenlinhlinh:
@stephan271 could we merge this one? I am going to create another PR for supporting multiple formats later when I have time.

PR #815, opened 2021-08-17
Lazy loading

nguyenlinhlinh:
## Description
This PR tries to reduce the bundle size and lazy load modules.
### Optimizing bundle size

How we import modules has big impact on the bundle size. One example is that we need to use `batch-card` component in `LayoutModule`. If we import the `DatasetsModule` into `LayoutModule` it will increase the size of `main.js` by ~1MB comparing with only importing the `BatchCardModule`. I tried very hard to reduce the bundle size however it is impossible to reduce the size of the shared module because it uses `exceljs` which is more than 1 MB in production build.
After working with this PR I learned that:
- It is better to have one modules per component instead of having multiple components in one module. 
- If we just want to use a single component in a module with multiple components then it is better to creating a module for that component and just import it. 
- Unused imported modules also increasing the bundle size so we should remove them (It is sad that angular doesn't have warning for these).

### Lazy loading

There might be some questions about why I created  `lazy` folder (solution 1) instead of just adding routing files to modules across the app (solution 2). I tried both solution and they both have advantages and disadvantages but I think solution one is better because:
- It is easier to have an overview on the routing tree, on which routes we have
- It is easier to avoid accidentally importing routing modules in other modules (which can causing bugs if we do)
## Changes:
- Lazy load modules
- Optimized bundle size by removing unused import and importing individual components
Bundle size for production build after optimization 
Screenshot 2021-08-17 at 11 59 20
## For reviewers
- There are some routes that MAXIV doesn't have or doesn't data for them to show up properly so please try them out to see if they work as expected. These are /instruments(child paths) and /logbooks (child paths)
- In the `*.routing.module.ts` I added comments on some routes that we might not need anymore. If you don't have any objection I am thinking on removing them.
- Are these paths still relevant or we should remove them? `help/ingestManual, help/ingestManual, help/ingestManual, SciCatGettingStartedSLSSummary `

- [ ] Docs updated?

### DIFF
--- a/src/app/_layout/layout.module.ts
+++ b/src/app/_layout/layout.module.ts
@@ -7,12 +7,12 @@ import { MatIconModule } from "@angular/material/icon";
 import { MatMenuModule } from "@angular/material/menu";
 import { MatToolbarModule } from "@angular/material/toolbar";
 import { RouterModule } from "@angular/router";
-import { DatasetsModule } from "datasets/datasets.module";
 import { AnonymousLayoutComponent } from "./anonymous-layout/anonymous-layout.component";
 import { AppLayoutComponent } from "./app-layout/app-layout.component";
 import { LoginHeaderComponent } from "./login-header/login-header.component";
 import { LoginLayoutComponent } from "./login-layout/login-layout.component";
-import { SharedCatanieModule } from "shared/shared.module";
+import { BatchCardModule } from "datasets/batch-card/batch-card.module";
+import { BreadcrumbModule } from "shared/modules/breadcrumb/breadcrumb.module";
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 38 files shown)


stephan271:
"Are these paths still relevant or we should remove them? help/ingestManual, help/ingestManual, help/ingestManual, SciCatGettingStartedSLSSummary "

These are URLs to helper documentation. Where do you want to remove them. They

stephan271 (commented):
LGTM

nguyenlinhlinh:
@stephan271 you closed this by accident?

stephan271:
Yes, probbaly. Dont know how, however, but yes, that was not my intention. Do I need to do anything to repair this ?

nguyenlinhlinh:
@stephan271 :) No problem I just reopen it.

stephan271:
Thanks !

nitrosx:
@nguyenlinhlinh : do you need help on this PR?

nguyenlinhlinh:
@nitrosx I am done with this PR. I just have to make some test case to pass the coverage test. Since it effects many places it could be really helpful if you could try it out. Stephan has reviewed it but I hope to have at least 2 reviewer just to make sure.

nitrosx:
OK. I will review it by tomorrow morning and let you know

nguyenlinhlinh:
Any objection to merging this one  :) @nitrosx ?

henrikjohansson712 (approved):
Looks good!

PR #824, opened 2021-08-27
ALS Build

dylanmcreynolds:
## Description

- Fix ALS Configurations for Catanie
<del>- Add a github action job for building and publishing to ghrc.io</del>

## Motivation

Cleaning up and fixing ALS dev, staging and production builds.

Plus, I added a github action job creating a single catanie container that can be built and used for any catanie configuration. This is now possible given the merging of #774 from @minottic. The new job builds and publishes to github's ghrc.io. I am not 100% sold on this idea...putting it here provides build that can be used by everyone. I know that dacat exists in the docker registry, but I'm not sure how often it's being updated. Here you can see that this branch is being built and published on my fork: https://github.com/dylanmcreynolds/catanie/pkgs/container/catanie


dylanmcreynolds:
I have removed the ghcr addition. I'll probably do that in a private fork.

Issue #808, opened 2021-07-28
Fix/Add sharing of datasets


bolmsten:
## Use Case

There exist multiple use cases when it comes to sharing of data, they range from sharing all data on a proposal for a large group, to sharing a single dataset with an individual. 

It is proposed to start with the use case of sharing a number of dataset with one or more people. A PI or Co-proposer that has access rights to the data can select one or more users to get access to a selected number of datasets by adding their email addresses. It should also be possible to revoke access given to a user.

Different sites will have different requirements when it comes to who is allowed to share data, this could be fixed by leveraging the policy system in scicat.

### Proposed solution

Some functionality already exist in SciCat that tries to cater to the above mentioned use case, it is however not in a production ready state and is missing some features. 

As seen below a user can select a number of datasets and through the shopping cart select share, this will prompt the user to fill in a email address for an already registered user in SciCat. If the user does not have a registered account a message should be shown informing of this. It should also be possible to add multiple people in the modal shown.

<img width="1434" alt="Screenshot 2021-07-28 at 13 29 52" src="https://user-images.githubusercontent.com/6403388/127315336-d894e2c7-8e18-4576-bf14-69ec764f756b.png">

In the dataset view a user can se who has been shared a specific dataset and has the possibility to remove a user.

<img width="1436" alt="Screenshot 2021-07-28 at 15 27 38" src="https://user-images.githubusercontent.com/6403388/127330812-a1126092-dd0b-4595-beb8-fc92af3d31c5.png">

### Technical

Currently access and owner groups exist to control who can see what data, one option would be to extend this functionality and create groups on the fly when sharing datasets. This group would than be added to the concerned datasets under shared group. This solution would however require a significant amount of development as new GUI and logic would need to be developed for the handling of groups.

Another option would be to add selected users directly on the dataset, this would eliminate the need for groups and make for a simplified workflow. A user can than visit a dataset and remove a user if needed.

----- Add more information here about what is currently possible to do in catamel and the best technical way forward ----

Each site would need to handle what it means for a user to be given access to a dataset in there specific infrastructure. This becomes important as users would want to access data through other systems than scicat.

### Open questions

How can we let users select people to share with without breaking GDPR regarding sharing of user data?

Can we add people instead of groups directly on datasets without major refactoring?

dylanmcreynolds:
I think we should remove the current UI for sharing until a working implementation is built. It's fairly misleading right now.

nitrosx:
Addressed with PR #838

PR #838, opened 2021-09-13
Feature/share datasets

henrikjohansson712:
## Description

Makes it possible to share datasets with other SciCat users through their email. This is done through the batch view, where a user can type in one or more email addresses and share the selected datasets with these users. If an email address isn't connected to a current SciCat user, an error is thrown. Once a dataset has been shared with a user, the username of this user is listed in the "Shared With" filed in the dataset detail view. The dataset can then be unshared by clicking the "X" next to the username.

Further things to add to this feature is probably a limitation on which proposal members that should be able to share and unshare a dataset.

Depends on catamel [#482](https://github.com/SciCatProject/catamel/pull/482).

## Motivation:

See issue #808

## Changes:

* Add share button and dialog to batch view

- [ ] Docs updated?
- [x] Requires update of catamel API?


nguyenlinhlinh:
I am trying this feature out by checking out this branch in catanie and catamel. Do I need to do something more in order for this to work? I get this error "**The email address is not connected to a SciCat user**" eventhough the email I use is a Scicat user. The error I get from the backend is "**GET /api/v3/UserIdentities/findOne: Error: Unknown "UserIdentity" id "undefined**"

henrikjohansson712:
@nguyenlinhlinh Shouldn't need anything else, it's just that right now the email address needs to be an exact match of what is stored in the database. So it might be that the email address in the database has some uppercase letters.

nguyenlinhlinh:
@henrikjohansson712: Thank you for working with this.
I was able to find some of our test accounts in AD only and sometimes with the same test account (that was found before) I got the error that the email address is not connected. With the real user accounts it doesn't work but I assume that something is wrong in our configuration unless other facilities also have the same problem. 

I notice that after sharing, the dataset get a field Share With: 2054, 3749 which I don't know where they come from but I personally prefer to see the usernames as you showed in the meeting instead of number, since it is easier for the user to figure out which user it is at removing. 

Is it possible to add a configuration variable so that we can enable or disable this feature. We need to figure out how to deal with file access before we can allow user to use this feature. Nothing advance just a variable something like `sharingEnabled = true or false`  and then on the share button in batch view can be disabled depend on the variable `<button disable=!sharingEnabled >Share<button>`.

henrikjohansson712:
@nguyenlinhlinh 
> I notice that after sharing, the dataset get a field Share With: 2054, 3749 which I don't know where they come from but I personally prefer to see the usernames as you showed in the meeting instead of number, since it is easier for the user to figure out which user it is at removing.

Thought it would be better to discuss this in the catamel pr, see: https://github.com/SciCatProject/catamel/pull/482#issuecomment-919179513

> Is it possible to add a configuration variable so that we can enable or disable this feature. We need to figure out how to deal with file access before we can allow user to use this feature. Nothing advance just a variable something like sharingEnabled = true or false and then on the share button in batch view can be disabled depend on the variable <button disable=!sharingEnabled >Share<button>.

You're right, that's probably a good idea. Thanks!

nitrosx:
@henrikjohansson712: I just try to share one dataset with myself and when I view it I see the following:
Screenshot 2021-10-01 at 10 54 29

I was expecting to see my user name or email.

nitrosx:
Now I shared also with @henrikjohansson712, and his username/email is shared correctly
Screenshot 2021-10-01 at 11 01 25

nitrosx:
I checked in the database and here the dataset has the following info under sharedWith:
`  
"sharedWith": [
    "1ddc156f-0d46-3297-b03d-db47a68b03a7",
    "henrikjohansson"
  ],
`

@henrikjohansson712: should we prevent sharing with myself? Also, why when I shared with myself, it stored a uuid?

nitrosx:
I would like to see the correct username or email listed under "Shared With" in the details section of the dataset page.

henrikjohansson712:
@nitrosx Looks the `UserIdentity.externalId` field is a uuid for a few of our users. I guess it's better to use another field to populate the `Dataset.sharedWith` array, like `UserIdentity.profile.username`, since `UserIdentity.externalId` also caused issues for MAX IV.

@nguyenlinhlinh Would using `UserIdentity.profile.username` work for you?

nguyenlinhlinh:
> @nguyenlinhlinh Would using `UserIdentity.profile.username` work for you?
With username our shows up like this "Linh Nguyen". It is ok but if you use this parameter for distinguish between user it might not work for us since this parameter it is not unique.

nguyenlinhlinh (approved):

stephan271 (approved):

PR #846, opened 2021-09-24
add missing Effects

nguyenlinhlinh:
## Description

## Motivation

Some effects in state management was missing due to the refactoring in lazy loading PR. Could you check that the logbook action works in proposal and logbook page? we don't have logbook so I can't test it.

- [ ] Docs updated?


stephan271:
Dont have logbooks either, can Henrik test ?

PR #887, opened 2021-12-10
Retrieve dest

minottic:
## Description

Enable option to choose between different retrieve destinations. If specified, enable the selection of the path on the retrieve location. 

The option can be set by adding the `retrieveDestinations` option to the environment.ts file This is optional, and there is no need to add an empty list in case of no use. E.g.:
```
  retrieveDestinations: [{option: "PSI", location: "/home/out"}, {option:"CSCS (Testphase)"}]
```

Expected behaviour when enabling `retrieveDestinations`: 
* The user selects the option from a dropdown list, if the `location` field is non-empty for the selected option, a new field is shown, where she can select the location path, relative to the one in the `retrieveDestinations[i].location`.
* if no `location`, no further field will be shown
* two new fields are added in the job model, under `jobParams.location` and `jobParams.option`. If there is no `location in the `retrieveDestinations` option selected by the user, no  `jobParams.location` field is added to Job.

Below is an example:

1a. image
2a. image
3a. image
1b. image

and the corresponding results stored in the DB:
a. image
b.image

## Motivation

At PSI we need to support such a use-case. Users will be able to retrieve data to PSI or via email or to cscs

## Changes:

* archiving-service-spec -> change the archiving logic to store two more info in the job model (option and location)
* dataset-table-actions -> make use of retrieval option when selecting datasets in dataset table
* batch-view  -> make use of retrieval option when adding to cart
* dialog-component -> enable the user to choose between different options

- [ ] Docs updated?

### DIFF
--- a/src/app/app-config.module.ts
+++ b/src/app/app-config.module.ts
@@ -10,6 +10,11 @@ export class OAuth2Endpoint {
   authURL= "";
 }
 
+export class RetrieveDestinations {
+  option = "";
+  location?: string | null = null;
+}
+
 export class AppConfig {
   lbBaseURL = "";
   externalAuthEndpoint: string | null = null;
@@ -57,6 +62,7 @@ export class AppConfig {
   userNamePromptEnabled = false;
   loginFormEnabled = true;
   oAuth2Endpoints: OAuth2Endpoint[] = [];
+  retrieveDestinations?: RetrieveDestinations[] = [];
 }
 
 export const APP_DI_CONFIG: AppConfig = {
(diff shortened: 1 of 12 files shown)


minottic:
@stephan271 could you please delete the branch [feature/ChooseRetriveDestination](https://github.com/SciCatProject/frontend/tree/feature/ChooseRetriveDestination) once this PR is merged? The changes in that branch are the same as this one, but with a less tidy commit history

stephan271 (approved):

Issue #848, opened 2021-09-28
Add child routes for dataset details tabs


stephan271:
## Issue Name
ChildRoutesDatasetDetails

### Summary
When switching between Tabs in the dataset details view the URL should reflect this state change by adding child routes

### Extra Details

Prerequisite for fixing #847

nguyenlinhlinh:
you mean something like this https://scicat-test.maxiv.lu.se/datasets/[pid]/datafiles, https://scicat-test.maxiv.lu.se//datasets/20.500.12269%2F353bb674-1bbf-4829-9cdb-e4f1b0d58142/reduce?

nguyenlinhlinh:
WIP

stephan271:
Thanks for taking this up !

PR #888, opened 2021-12-13
Child routes

nguyenlinhlinh:
## Description
Add child routes for tabs in `dataset-details-dashboard.component.ts`.
See #848 for more information
## Motivation
See #848 
## Fixes:
- Add child routes
- Refactor the code in `dataset-details-dashboard.component.ts`. Delegate the logic parts to sub components in each tab
## For reviewers
- I want to disable the routes if  the value for a give feature is disabled in `environment.ts` for example route `.../logbook` should not exists if `logbookEnabled = false`. However I can't find out how to import the appConfig to `src/app/app-routing/lazy/dataset-details-dashboard-routing/dataset-details-dashboard.routing.module.ts`. Please let me know if you know how to do that.
- Please check that the functions supported on each tab works properly. We don't have all the supported features so I am unable to test all tabs
- Check that tabs for unused features don't show up and tabs for public dataset and private dataset are showed up or hided correctly. 

- [ ] Docs updated?


stephan271:
Hi Linh, checked the tabs and all looks nice here, thanks. Concerning your environment question: I am not sure if I understand the question correctly, but maybe this link helps ? https://stackoverflow.com/questions/51056704/configure-angular-routes-in-environment

nguyenlinhlinh:
@stephan271 Thank for the link. The link is what I am searching for even though my initial idea was not the same. Because I didn't know what to search for in Angular so I though I could somehow filter the list of routes using appConfig. The Stackoverflow solution looks better.

stephan271:
Glad I could help :-)

stephan271:
The findOne call on the dataset still fetches the full information ({"where":{"pid":"20.500.11935/3ab12715-2ae5-4a8e-9c38-c1d2375bd528"},"include":[{"relation":"origdatablocks"},{"relation":"datablocks"},{"relation":"attachments"}]} already when entering the dataset details overview. Could one also load the data lazily, i.e. add the {"relation":"origdatablocks"} etc only when the corresponding tab is selected ?

nguyenlinhlinh:
@stephan271 It is possible to load the data on demand. Since the task is about child routes I just make changes to be able to create child routes and preserve the logic for handling data untouched. But I can make changes so that the data is loaded lazily.

stephan271:
For us being able to load the data lazily is quite critical, because otherwise the browser starts to become unresponsive for datasets with many files. But doing this in a second PR would of course also be good, in fact I would recommend doing it in 2 steps, i.e. in a 2nd PR.

nguyenlinhlinh:
Ok, I will work on the second PR.

stephan271 (approved):

henrikjohansson712 (approved):

PR #899, opened 2022-01-21
Chore/user view model selectors

henrikjohansson712:
## Description

Replaces individual selectors in the user module with view model selectors. Plan is to do this for every component in the app.

## Motivation

Each view will now just make a single call to the store to get data. Easier to work with and easier to test as there will only be one async variable in each view.

## Changes:

* Create selectLoginPageViewModel and selectUserSettingsPageViewModel selectors
* Replace individual selectors in login component with selectLoginPageViewModel selector
* Replace individual selectors in user settings component with selectUserSettingsPageViewModel selector
* Remove unused onSubmit method from user settings component

- [ ] Docs updated?

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -1054,10 +1054,8 @@ exports[`stricter compilation`] = {
     "src/app/users/login/login.component.spec.ts:87531066": [
       [160, 8, 71, "Expected 0 arguments, but got 1.", "3870196511"]
     ],
-    "src/app/users/user-settings/user-settings.component.spec.ts:2325442996": [
-      [95, 8, 32, "Expected 0 arguments, but got 1.", "1341113725"],
-      [97, 47, 30, "Expected 0 arguments, but got 1.", "2329020575"],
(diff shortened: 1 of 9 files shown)


nguyenlinhlinh (commented):
All looks good except that the user icon is replaced with "no photo". Is this on purpose? 
Screenshot 2022-01-31 at 12 38 20

henrikjohansson712:
It worked with our users. Shows the profile photo if it exists, if not, it shows the user icon. Do you have the key "thumbnailPhoto" in userIdentity.profile, or is it stored somewhere else, or named differently?

nguyenlinhlinh:
@henrikjohansson712 It works on our deployment. We can see the icon there but not on this PR. Our user data looks like this
Screenshot 2022-01-31 at 13 23 59

henrikjohansson712:
@nguyenlinhlinh Added an additional check, try it now

nguyenlinhlinh (approved):

nitrosx (approved):

PR #915, opened 2022-02-14
Runtime configuration

henrikjohansson712:
## Description

This pr replaces the build time configs with runtime configs. The config can either be served from the backend API (currently at the */api/v3/config* endpoint) or added as a json file in */usr/share/nginx/html/assets* in the built docker image. For development purposes, a default *config.json* is already placed in the folder. This json file corresponds to the previous default *environment.ts* file in *src/environments*. I have added json files corresponding to the removed environments in each facility's CI folder, so that no configs are lost. These should be removed once you have set up run time configs in your deployment pipelines. 

Since all environments except for **default** and **production** have been removed, image files are no longer automatically replaced on build. These should instead be mounted into the */usr/share/nginx/html/assets/images* folder in the built docker image.

## Motivation

Make it easier for people to get SciCat up and running by supplying one official docker image, that can be easily configured by the facility itself.

## Changes:

* Remove all environments except for **default** and **production**
* Replace build time configs using app config module with run time configs using app config service
* Update Dockerfile to use multi stage build
* Remove everything related to PWA

- [ ] Docs updated?


nitrosx (changes requested):
The code looks good.
What is the best way to test it?
Do we have a configuration that can be used to run it locally?

henrikjohansson712:
@nitrosx You can run it by checking out the branch, run `npm ci` and then `npm start`.

nitrosx:
I was able to run it successfully following your directions.
The login page is shown without any background image. Is that correct?

nitrosx (approved):

dylanmcreynolds (approved):
This is so very good! I didn't give it a line-by line read, but I tested it in my development environment. I look forward to the merge.

henrikjohansson712:
@nguyenlinhlinh  @minottic

nguyenlinhlinh:
@henrikjohansson712 Sorry for the delay. I will check this out today.

nguyenlinhlinh (approved):

LAShemilt:
Works well for us , I just tried this simply in our docker-compose script for spinning up a local dev version of scicat. Everything works really well. Thanks for going to the effort to preserve the environments in the CI/<InstituteName> dir.

minottic (approved):
sorry for the dealy and indeed works very nicely when testing! thank you for the change

dylanmcreynolds:
When this is merged, I will likely create a release according to the new release process that @nitrosx documented.

stephan271:
:+1:

PR #945, opened 2022-04-06
Hotfix/token prefix config

henrikjohansson712:
## Description

New backend uses Bearer tokens. This PR adds an `accessTokenPrefix` config property that should be set to "Bearer " when using the new backend and just an empty string when using the old backend.

## Motivation

Make Frontend compatible with both backends.

## Changes:

* Add `accessTokenPrefix` config prop to app config service
* Implement logic to set prefix in user effects

- [ ] Docs updated?


nitrosx:
@nguyenlinhlinh , @dylanmcreynolds and @minottic: could you please review this PR and approve it if it is ready?

nguyenlinhlinh (approved):

minottic (approved):
LGTM. Only comment: is there a reason why whenever there is a new config option we make it mandatory to specify it? Couldn't we fallback to "" if not specified? I think that makes it easier to upgrade to newer versions without needing to change the config file (or needing to know all the config file+ effects changes)

nitrosx:
@henrikjohansson712 Could you please address the question from @minottic ?
@dylanmcreynolds Could you please review and approve this PR?

henrikjohansson712:
@minottic Because we're using Typescript. If we don't specify the variable, then it becomes pretty much impossible to access, which is annoying.

minottic:
ah ok, thanks. But I am pretty sure I've done that in here: #887 , maybe then it's not the best way?

PR #974, opened 2022-06-10
Additinal CONTAINS_STRING option for characteristic filter

linupi:
## Description

In https://github.com/SciCatProject/backend/pull/680 we propose a filter option for partially matching strings and here we added the corresponding frontend modifications.

## Motivation

imagine you'd have the following metadata for a dataset

```json
"scientificMetadata": {
        "localContact": "John Doe",
        "experimentalists": [
                "John Doe",
                "Max Mustermann"
         ]
```

now we would like to filter all datasts where `John` participated in the measurement (or was local contact respectivly). So far this type of query was not possible using the options offered when clicking `add condition`  in the search box as there was only an option to exactly match complete strings and no option for partial string matching or matching strings in lists.

## Changes:

new_contains_string_option

condition_label

- [ ] Docs updated?
- [x] Requires update of SciCat backend API?

### DIFF
--- a/src/app/datasets/datasets-filter/datasets-filter.component.html
+++ b/src/app/datasets/datasets-filter/datasets-filter.component.html
@@ -185,6 +185,9 @@
           <ng-container *ngSwitchCase="'EQUAL_TO_STRING'">
             &nbsp;=&nbsp;
           </ng-container>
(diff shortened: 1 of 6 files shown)


bpedersen2:
rebase onto current HEAD and created a fresh pull request #1342

PR #990, opened 2022-10-31
Logbook access and proposal display

nitrosx:
## Description

This PR add the following new features:
- show the proposal name even if it is not present. This feature is controlled by option datasetDetailsShowMissingProposalId in config file
- provide logbook access to users that part of dataset access group.
 
## Motivation
When data is ingest automatically, we do not know if the proposal associated has already been created. The system allows to create a dataset with a proposal id that does not exists in scicat. If that happens we need to be able to display it and follow up with the proper actions.
Logbooks where not accessible to users belonging to access groups. We are going to use the access groups for beam lines scientists and other facility experts and we would like to grant access to every details of the dataset using access groups.

- [x] Docs updated?
- [x] Requires update of SciCat backend API?

### DIFF
--- a/src/app/_layout/app-header/app-header.component.html
+++ b/src/app/_layout/app-header/app-header.component.html
@@ -22,13 +22,14 @@
       </button>
     </ng-container>
 
-    <ng-container *ngIf="config.logbookEnabled">
+<!--    <ng-container *ngIf="config.logbookEnabled">
       <button mat-menu-item routerLink="/logbooks/">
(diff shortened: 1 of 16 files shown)


minottic (commented):
LGTM. I would just remove the commented lines and I am fine merging. Out of interest, how do you update the proposal when the proposal number is available?

nitrosx:
@minottic : All comments removed!!!

minottic (approved):

PR #999, opened 2022-11-22
fix: remove printing oAuth endoints

jekabs-karklins:
## Description

This PR removes debug message which prints out all oAuth endpoints

## Motivation

There should be no debugged info printed out

## Changes:

Login screen


minottic (approved):
thanks! I was doing the same thing!

PR #1029, opened 2023-03-02
refactor: login screen refactored to show external login first

martin-trajanovski:
## Description

Login screen refactored to show external login options first (if any available) with tabular view. Another option for login is local which means login with functional account. Added some e2e tests for it.


nitrosx (approved):
I've notice that you added support for only one OIDC service?
Have you thought about allowing multiple of them?

martin-trajanovski:
> I've notice that you added support for only one OIDC service? Have you thought about allowing multiple of them?

@nitrosx Not sure what do you mean by "only OIDC" service. I think that's configurable and you can add as many services as you want. Or maybe I misunderstand you somehow

PR #1038, opened 2023-03-19
fix: created proposal-logbook component & fixed few bugs

Junjiequan:
## Description

- feat: created new logbook component for proposal page
- fix: fixed logbook doesn't display issue when proposal have no dataset.
- fix: fixed logbook filter search bar bug  - when search query is empty, it returns event method instead of string value.

## Motivation
New proposal logbook component makes api call based on only proposal id.
Using mixed logic of dataset pid & proposal id based logbook call increases complexity of the component and makes the code hard to maintain.

- [ ] Docs updated?

### DIFF
--- a/src/app/logbooks/logbooks-dashboard/logbooks-dashboard.component.ts
+++ b/src/app/logbooks/logbooks-dashboard/logbooks-dashboard.component.ts
@@ -49,9 +49,6 @@ export class LogbooksDashboardComponent
   implements OnInit, OnDestroy, AfterViewChecked {
   vm$ = this.store.select(selectLogbooksDashboardPageViewModel);
 
-  @Input() proposal_logbook: LogbookData;
-  @Input() proposal_dataset: TableData;
-
   dataset: Dataset | undefined = undefined;
   appConfig = this.appConfigService.getConfig();
 
@@ -68,7 +65,6 @@ export class LogbooksDashboardComponent
 
   applyRouterState(pid: string, filters: LogbookFilters) {
     console.log("Rerouting to Dataset Logbook");
-    if (this.proposal_logbook) return;
 
     this.router.navigate(["/datasets", pid, "logbook"], {
       queryParams: { args: JSON.stringify(filters) },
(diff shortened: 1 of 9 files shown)


src/app/logbooks/logbooks-dashboard/logbooks-dashboard.component.ts
```diff
@@ -68,20 +63,17 @@ export class LogbooksDashboardComponent
...
   }
 
   onTextSearchChange(pid: string, query: string) {
-    const queryText = query.trim();
-    
-    if(queryText.length > 0){
-      this.store.dispatch(setTextFilterAction({ textSearch: queryText }));
-      this.store.dispatch(fetchDatasetLogbookAction({ pid }));
-    }
+    // TODO: query.length return undefined when it is empty
```
martin-trajanovski, line 73: Is this a TODO or just a NOTE that explains why we need to check `query.length`

src/app/proposals/proposal-logbook/proposal-logbook.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  @Input() logbook: LogbookData;
+
+  constructor(
+    public appConfigService: AppConfigService,
+    private cdRef: ChangeDetectorRef,
+    private store: Store,
+  ) {}
+
+  onTextSearchChange(proposalId: string, query: string ) {
+    // TODO: query.length return undefined when it is empty
```
martin-trajanovski, line 56: The same goes here

martin-trajanovski (approved):
Just small comment otherwise it looks good to me

src/app/logbooks/logbooks-dashboard/logbooks-dashboard.component.ts
Junjiequan: Yes, It's just an explanation here, I've included an If condition to handle the issue, but its kinda temporary solution. We might need to dig deep into in the cause later.

nitrosx (approved):

src/app/logbooks/logbooks-dashboard/logbooks-dashboard.component.ts
nitrosx: Let's keep it

src/app/proposals/proposal-logbook/proposal-logbook.component.ts
nitrosx: Same as above

PR #1048, opened 2023-04-13
PID Search for everyone

TheMeinerLP:
## Description

Add for archive manager a special pid search

## Motivation

I am the son of Michael Kallmeier-Glanz he showed me the sciCat. In some points the workflow for him are non productive.
About that reason i will try to add some nice features that helps him.
## Fixes:

- None

## Changes:

- Added a PID filter for dataset view

- [ ] Docs updated?
- [X] Toggle added for new features?

## Other 
I need littlebit help/informations how can i hide the field for non archive manager users.
And why the filter is not applying to the PID and the auto completion is empty.

### DIFF
--- a/CI/ESS/e2e/cypress.json
+++ b/CI/ESS/e2e/cypress.json
@@ -8,6 +8,8 @@
   "password": "am2jf70TPNZsSan",
   "secondaryUsername": "archiveManager",
   "secondaryPassword": "aman",
+  "guestUsername":"user1",
+  "guestPassword":"a609316768619f154ef58db4d847b75e",
   "defaultCommandTimeout": 10000,
   "retries": 1
 }
--- a/CI/ESS/e2e/cypress.nestjs.json
+++ b/CI/ESS/e2e/cypress.nestjs.json
@@ -8,6 +8,8 @@
   "password": "am2jf70TPNZsSan",
   "secondaryUsername": "archiveManager",
   "secondaryPassword": "aman",
+  "guestUsername":"user1",
+  "guestPassword":"a609316768619f154ef58db4d847b75e",
   "defaultCommandTimeout": 10000,
   "retries": 1
 }
--- a/CI/PSI/config.development.json
+++ b/CI/PSI/config.development.json
@@ -118,5 +118,6 @@
   "scienceSearchUnitsEnabled": false,
   "metadataStructure": "",
   "loginFormEnabled": true,
-  "oAuth2Endpoints": []
+  "oAuth2Endpoints": [],
+  "prefix": "20.500.11935"
 }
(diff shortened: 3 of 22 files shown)


stephan271:
I think such a feature should be implemented for all users, not just archiveManagers.
Can you provide a screenshot /Mockup of what you plan to do and how it looks to the users ?

TheMeinerLP:
Of course i can provide it. 
image
The code for the input is already here in the PR. But the completion and the search function of it self is not working.

stephan271:
I think PID should not be treated as a "facet", as the other fields are, because PIDs are different for each dataset. I Think the main use case is that a user has copied a PID from somewhere and now wants to find it, so the search expression would be a simple equality check on the PID field. There is no need need to calculate any facet counts. Is this what you/Michael have in mind ?

TheMeinerLP:
As other idea i already have to add a small icon to the pid id on general Information to copy the id more faster. But when then happend i will split into extra pr 
image

TheMeinerLP:
> I think PID should not be treated as a "facet", as the other fields are, because PIDs are different for each dataset. I Think the main use case is that a user has copied a PID from somewhere and now wants to find it, so the search expression would be a simple equality check on the PID field. There is no need need to calculate any facet counts. Is this what you/Michael have in mind ?

Yeah thats exactly what was in my mind.

TheMeinerLP:
How can finish implement for you. Means what needs be removed/replaced with other code?

TheMeinerLP:
image
Final solution after some background writing with @stephan271.
Thanks for that help!

Now users can be post there PID into a field and press enter or the button inside of the text field to get direct to there dataset.

Also the system checks if the a PREFIX of Company x included. 
Mean when a use forgot like `20.500.11935/VALUE` of the selection. The tool will be added automaticlly into the request.

nitrosx:
My apologies for jumping in last minute:
Have you considered the case where I have a partial PID? Would partial match be also useful?
We can add this as a following PR.

nitrosx (approved):

TheMeinerLP:
I think thats also a good point. Can you deliver me a example i will open another PR for that 😄 
First on i will fix the github actions

stephan271:
Just a comment to the partial PID reqiest: partial PID (substring) search will need another API endpoint using where clauses with regular expressions. However I am not sure that this is even possible on fields of type ObjectId. Also one would need to take into account that multiple datasets can be returned, therefore one would need to display the result table first before showing the dataset details.

nitrosx:
@stephan271 : I agree with you. Partial match on id is not optimal.
When we have migrated to the BEv4.x, datasets have their id duplicated in a _pid_ field which is a simple string, so we can do it than.
I am going to approve this PR and we will open a new PR for the partial match when the time is right

nitrosx (approved):

PR #1050, opened 2023-04-18
fix: files page does not render data

Junjiequan:
## Description

Fetch data for files dashboard from new endpoint

## Motivation

Files page does not render data.

## Fixes:
image

To

image

## Changes:

* changes made

- [ ] Docs updated?


nitrosx (commented):
I don't think that creating a specific if statement for the files in the dataservice is a good idea.
This opens the door to single endpoint customization that is dangerous.
That said, I'm not sure if there is a better solution.

nitrosx (approved):

PR #1051, opened 2023-04-18
Change dataset details page PID text to a input field with copy button

TheMeinerLP:
## Description

On the detail page of a dataset is a PID. Normaly the user needs to select manually the PID. This PR adds a input field with simple copy option. 

## Motivation

To deliver more easy a dataset PID add a UX improvement.

## Fixes:

* Nothing

## Changes:

* UX improvement

- [ ] Docs updated?
- [X] Toggle added for new features?

Before:
image

After:
image

### DIFF
--- a/src/app/datasets/dataset-detail/dataset-detail.component.html
+++ b/src/app/datasets/dataset-detail/dataset-detail.component.html
@@ -68,7 +68,7 @@
               <th>PID</th>
               <td>
                 <mat-form-field class="full-width">
-                  <input matInput type="text" [value]="dataset.pid">
+                  <input matInput type="text" [value]="dataset.pid" disabled>
                   <button matSuffix mat-icon-button aria-label="copy" (click)="copyIntoClipboard(dataset.pid)">
                     <mat-icon>content_copy</mat-icon>
                   </button>
--- a/src/app/datasets/dataset-detail/dataset-detail.component.html
+++ b/src/app/datasets/dataset-detail/dataset-detail.component.html
@@ -66,7 +66,14 @@
             </tr>
             <tr>
               <th>PID</th>
-              <td>{{ dataset.pid }}</td>
+              <td>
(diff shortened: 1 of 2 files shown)


src/app/datasets/dataset-detail/dataset-detail.component.html
```diff
@@ -66,7 +66,14 @@
             </tr>
             <tr>
               <th>PID</th>
-              <td>{{ dataset.pid }}</td>
+              <td>
+                <mat-form-field class="full-width">
```
nitrosx, line 70: Instead of creating a full functional text box, can we just add the copy button, like is done under user settings with the [scicat token](https://github.com/SciCatProject/frontend/blob/757d07978013d7161aa54c6f6b2513f728c427c0/src/app/users/user-settings/user-settings.component.html#L52)

nitrosx (changes requested):
Check the scicat token copy under the user settings page. 
We should implement it as it is done there

src/app/datasets/dataset-detail/dataset-detail.component.html
Junjiequan: From the best practice point of view, especially if you are an accessibility enthusiast, I wouldn't recommend to use onClick in a span to make it a button.
But in our case, I think it is completely fine to follow the pattern what we already have for consistency, and you can add an aria attribute if you want to.

TheMeinerLP: Okay i will check that from the settings page and improve the current code

minottic:
after the changes, I would add a test, similar to this: https://github.com/SciCatProject/frontend/blob/757d07978013d7161aa54c6f6b2513f728c427c0/src/app/users/user-settings/user-settings.component.spec.ts#L72-L91

nitrosx (approved):

PR #1052, opened 2023-04-25
fix: 401 unathorized when login into scicat

Junjiequan:
## Description

Separated local account login and ldap account login as two different login methods.  

## Motivation

Previously when attempting to use local login, it throws `401 unauthorized` regardless of whether or not  the login attempt is  success. 
This issue occurs because local login trigger both ldap login & local login in sequence. Even local login is valid, backend would still return rejection for ldap endpoint trigger.

## Fixes:

* Items added

## Changes:

https://user-images.githubusercontent.com/78078898/234282240-eadb717e-ff8b-400a-9a93-245f6c8a8249.mp4

- [ ] Docs updated?


src/app/state-management/effects/user.effects.ts
```diff
@@ -139,28 +137,31 @@ export class UserEffects {
...
 
-  loginRedirect$ = createEffect(() => {
-    return this.actions$.pipe(
-      ofType(fromActions.activeDirLoginFailedAction),
-      map(({ username, password, rememberMe, error }) =>
-        fromActions.funcLoginAction({ username, password, rememberMe, error })
-      )
-    );
-  });
+  // loginRedirect$ = createEffect(() => {
```
nitrosx, line 140: Let's remove this commented function

nitrosx (changes requested):

PR #1059, opened 2023-05-16
add dataset id to zip data

bpedersen2:
# Description

Add the dataset id as data pushed to the zip service

## Motivation

The dataset id can(together with the jwt that is already passed) be used to validate access to the data.

## Changes:

* add dataset id as hidden field

- [ -] Included for each change/fix?
- [ X] Passing? (Merge will not be approved unless this is checked) 
- [ -] Docs updated?
- [- ] New packages used/requires npm install? 
- [- ] Toggle added for new features?
- [- ] Requires update of SciCat backend API?

### DIFF
--- a/src/app/datasets/datafiles/datafiles.component.html
+++ b/src/app/datasets/datafiles/datafiles.component.html
@@ -49,6 +49,7 @@
       #downloadAllForm
     >
       <input type="hidden" name="jwt" [value]="jwt?.jwt" />
+      <input type="hidden" name="dataset" [value]="datasetPid"/>
       <input type="hidden" name="directory" [value]="sourcefolder" />
       <input
         *ngFor="let file of getAllFiles(); index as i"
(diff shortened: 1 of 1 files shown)


nitrosx (commented):
I wonder if this is the best way to provide authorization info to the zip service?
How does the zip service know if this user is authorized to download this dataset?

bpedersen2:
Yes , this is only one part. The zip service needs to do a call to the api/v3/ datasets/{pid} endpoint with the right token (in an OIDC setup this fails with  jwt token currently) and/or inspect the jwt token to  check if the user is allowed to download ( check access groups on dataset against jwt ).  

Probably the FE download should provide the same Authorization headers as any other scicat api calls to begin with, then the logic should be easier to implement. Still the  dataset should be helpful, as there are more fields in the dataset that may be needed to determine the path to the source files (e.g.sourceFolderHost).

nitrosx (approved):

nitrosx:
@bpedersen2: would you mind checking and resolving the conflicts?
I tried to merge your PR, but I'm not able to do it

bpedersen2:
Yes, I'll  rebase locally and push an updated branch.

PR #1061, opened 2023-05-16
Filter by PID enabling different matching conditions

minottic:
## Description

It enables to use PID field for filtering and returns a list of matching results before jumping to the details page. A toggle allows defining the condition to use when doing the search. Currently, three options are supported:

* equals
* contains
* startsWith

Please note that `contains` is intensive for the DB, so evaluate if to create an index on the db or speed it up by other means. 
The toggle is optional and defaults to `equals`. To set it explicitly, set the `pidSearchMethod` value in the config.json file

It looks like this: 

image

## Motivation

Users often want to filter by PID. Supporting regexes on this field will be useful when the PID format will encode some information from the facility. This will be possible when we will allow defining dataset PIDs from the outside at dataset creation. 

## Changes:

* PR #1048 reverted
* added action to dispatch PID filters

- [ ] Docs updated?


stephan271:
Fine for me to do 2 PRs

stephan271 (approved):

nitrosx (approved):

nitrosx (approved):

PR #1063, opened 2023-05-17
Pass right sorting column

minottic:
## Description

When using the column runNumber the sorting is buggy as it belongs to scientificMetadata in the be
## Motivation

When using the column runNumber the sorting is buggy as it belongs to scientificMetadata in the be

## Fixes:

* src/app/datasets/dataset-table/dataset-table.component.ts

- [ ] Docs updated?


nitrosx (approved):
I think we should review the dataset and maybe move the run number as a high level fields.
As it is right now, we rely on the user to add a metadata key "runNumber" which is prone to errors

PR #1067, opened 2023-05-24
add auth token to zip service

bpedersen2:
## Description

To allow the zip-service full evaluation of data access permissions the JWT seems not to be sufficient

## Motivation

In order to allow the zipservice access to the dataset api, the auth token is required, the jwt was not sufficient in this case.

Probably the use of a 'simple form' should also be replaced with an ajax post request, I'll look into this.

## Fixes:

*  add the auth token to the passed data

- [ - ] Included for each change/fix?
- [ X ] Passing? (Merge will not be approved unless this is checked) 
- [ - ] Docs updated?
- [ - ] New packages used/requires npm install? 
- [ - ] Toggle added for new features?
- [ - ] Requires update of SciCat backend API?

### DIFF
--- a/src/app/datasets/datafiles/datafiles.component.html
+++ b/src/app/datasets/datafiles/datafiles.component.html
@@ -48,6 +48,7 @@
       style="display: inline"
       #downloadAllForm
     >
+      <input type="hidden" name="auth_token" [value]="auth_token" />
       <input type="hidden" name="jwt" [value]="jwt?.jwt" />
       <input type="hidden" name="dataset" [value]="datasetPid"/>
       <input type="hidden" name="directory" [value]="sourcefolder" />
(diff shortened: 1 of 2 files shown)


bpedersen2:
the downloader can be found here: https://github.com/bpedersen2/mlzscicatdownloader

nitrosx (commented):
@bpedersen2 : I am wondering why you want to send the authorization token to the zip file?

bpedersen2:
> Have you checked if the new parameter impacts the official zip service? 

Should not have an impact, as the JWT is still passed the same way.

> Also, can we add a config parameter, so this feature can be disabled? I envision that some people do not want to use it. Thanks

Yes, that should be possible. But I fear the original downloader has the same problem.

nitrosx:
Thanks for taking the time to answer.
Please add the config parameter to enable/disable such feature and I will approved.

bpedersen2:
Feature flag add (defaults to false to be fully backwards compatible).

PR #1084, opened 2023-05-31
feat: add possibility to remove shared users from dataset

martin-trajanovski:
## Description

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/src/app/datasets/batch-view/batch-view.component.ts
+++ b/src/app/datasets/batch-view/batch-view.component.ts
@@ -8,6 +8,7 @@ import {
   clearBatchAction,
   prefillBatchAction,
   removeFromBatchAction,
+  storeBatchAction,
 } from "state-management/actions/datasets.actions";
 import { Dataset, Message, MessageType } from "state-management/models";
 import { showMessageAction } from "state-management/actions/user.actions";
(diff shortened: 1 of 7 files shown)


src/app/datasets/batch-view/batch-view.component.ts
```diff
@@ -121,13 +132,24 @@ export class BatchViewComponent implements OnInit, OnDestroy {
...
+        const datasetUpdatedBatch = this.datasetList.map((item) => ({
+          ...item,
+          sharedWith: result.users,
+        }));
+
+        this.clearBatch();
+        this.storeBatch(datasetUpdatedBatch);
+
+        const message = new Message(
+          result.users.length
```
nitrosx, line 146: how can you decide if a user has been added or removed by looking at the length of the users array?
shouldn't we compare it to the previous value?
Can you check if the message provided is consistent with the action?

nitrosx (changes requested):

src/app/datasets/batch-view/batch-view.component.ts
martin-trajanovski: This is only used when we want to remove all the users at once. There is a button called "Remove all" and then this message comes in place. If one user is removed or you modify the list of users, you just click "Share" again and it will show the default message.

martin-trajanovski: This is not ideal solution but for now it was the quickest one and easiest to implement.

nitrosx: Thanks for clarifying

nitrosx (approved):

PR #1107, opened 2023-06-15
fix login component for oidc callback listener

ddijken:
## Description

This listener on /login tries to get the parameters of access-token and user-id in an incorrect way and thus login does not succeed.

## Motivation

MAXIV prefers to use the /login listener rather than the /auth-callback listener (same function) because when login fails, the user remains on the login page rather than seeing the public datasets page.

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?


nitrosx:
@ddijken can you fix the linting issues and make sure that the tests passes?
Thanks

nitrosx (approved):

Issue #1135, opened 2023-07-21
Add possibility to use different images for the header and the login logo


fpotier:
### Summary
Atm, the frontend uses the asset named `site-logo.png` to display the facility's logo in the header bar and on the login page.
There is a hard coded behavior for the ESS to use another image for the header bar.

In the login HTML template:
```
<div class="login-container" fxLayoutAlign="center center">
...
          <img src="assets/images/site-logo.png" />
...
```

And in the header HTML template:
```
<ng-template [ngIf]="facility === 'ESS'" [ngIfElse]="otherFacility">
        <img class="logo" src="../../../assets/images/esslogo-white.png" />
</ng-template>
<ng-template #otherFacility>
        <img class="logo" src="../../../assets/images/site-logo.png" />
</ng-template>
...
```

I need to do the same thing for SOLEIL so I suggest to simply use two differents images instead of hard coding the behavior (facilities using the same image for both could then mount the  image on two different paths)

nitrosx:
I second that!!!

PR #1136, opened 2023-07-21
Fix proposal for #1135

fpotier:
See #1135


fpotier:
Note that this patch requires to update the deployment's config

nitrosx:
@fpotier would you be able to expand the functionality in the following way:

- If no configuration value is provided under the key "logo-icon", you use the assets/images/site-logo.png, otherwise you inject the value provide in configuration
- Update/Create documentation so we explain the configuration variable and value, and also we provide example on how to mount the logo image into the docker container.

Please let me know if it makes sense, or I should elaborate on it.

fpotier:
@nitrosx Should this new configuration key `logo-icon` contain the absolute path of the asset or we just assume it's a file in `assets/images`?

Where should I write the associated documentation?

nitrosx:
@fpotier let's assume that it leaves in `assets/images`.

You can find the documentation in repo https://github.com/SciCatProject/documentation.
Most likely goes under the Operator section.

nitrosx:
@fpotier please let me know if my comment makes sense

fpotier:
@nitrosx Yes it does, now it should work as you asked

nitrosx (approved):

PR #1171, opened 2023-08-21
Update angular and deps to 16.2.1

bpedersen2:
Still needs  debugging


bpedersen2:
Status: 
  + Unit tests are passing
  - Routing is broken....

bpedersen2:
E2E tests are failing ( seems like the backend is just not reachable, still investigating)
Locally it now works.

TODOS:
  - review the CSS changes for  all instituions
  -  verify other customizations

nitrosx:
@bpedersen2 have you checked if the accounts are correct? I mean the accounts info between frontend and backend.

bpedersen2:
> @bpedersen2 have you checked if the accounts are correct? I mean the accounts info between frontend and backend.

Yes, passwords were ok. But the it seems that the workflow used the old `cypress.json` config, I am retrying now with `cypress.nestjs.json` that uses different ports (#bc8c87f1).

bpedersen2:
Still TODO: review css 

I can present the current look during tomorrows meeting

nitrosx:
That's great!!!! 
Do you want me to merge this PR, or should we wait?

bpedersen2:
We should discuss it tomorrow, I think.

bpedersen2:
Cypress update was the culprit, the config file changed from json to ts/js, see https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-100

bpedersen2:
Should be OK to merge now

.github/workflows/test.yml
```diff
@@ -10,12 +10,12 @@ jobs:
...
     steps:  
       - name: Check out
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         
       - name: Set up Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version: 18.x
```
enthusiastio, line 18: It is advised to add the node version to the cache key. This is due to the fact that node package C++ addons could be compiled differently for different node versions. See here.
https://github.com/SciCatProject/frontend/pull/1179/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R29

enthusiastio (approved):
I have reviewed the node upgrade part here and it looks good, just a small comment

.github/workflows/test.yml
bpedersen2: Yes, a good idea. I'll do a followup

nitrosx (approved):

PR #1202, opened 2023-09-14
Bump prettier to 3.0.3 and fix warnings

bpedersen2:
- add a separate eslint step to the git hub workflow
 - adjust configuration
 - use same settings for prettier as in BE

Supersedes: #1193


src/app/app-routing/app-routing.module.ts
```diff
@@ -1,5 +1,5 @@
 import { NgModule } from "@angular/core";
-import { RouterModule, Routes, ExtraOptions} from "@angular/router";
+import { RouterModule, Routes, ExtraOptions } from "@angular/router";
```
dylanmcreynolds, line 2: Should we fix this wawrning?

dylanmcreynolds (dismissed):
I like this. 

I didn't do a line-by-line review, but in a couple places I noticed that we now have some ESLint errors, which should probably get fixed in this PR.

Thanks!

src/app/app-routing/app-routing.module.ts
bpedersen2: definitely, but I would like to keep this  change to the auto-formatting-only fixes and do a followup instead. This way it is easier to review.

nitrosx:
@bpedersen2 could you please resolve the issues? Thanks

nitrosx (approved):

bpedersen2:
@nitrosx : Merge if you are fine with this.

PR #1274, opened 2023-10-17
fix: redirect to logoutURL if provided from the backend after logout

martin-trajanovski:
## Description

Redirect to the logoutURL if it is provided from the backend after user logs out.


Junjiequan (approved):
Looks good to me

PR #1275, opened 2023-10-19
fix: change backend start command after build config change

martin-trajanovski:
## Description

This PR aims to fix the startup of backend container after build config is changed.

It depends on this ticket: https://github.com/SciCatProject/scicat-backend-next/pull/823


bpedersen2 (approved):
Merge after BE change

PR #1321, opened 2023-11-14
e2e improvement

bpedersen2:
Start docker-compose as real background service and add improved log output on failure


bpedersen2:
Seems like elasticsearch does fail sometimes, see the new detailed  docker logs output

Junjiequan (approved):
It's definitely nice idea to include docker logs!

PR #1373, opened 2024-01-17
Added flag to conditionally display unit list in metadata editing table.

Junjiequan:
## Description

Added `metadataEditingUnitListDisabled` flag to conditionally display available unit list in the metadata editing table.

## Motivation

The unit list is hard coded and shows only a limited unit list, which in some circumstances can give the user a false perception that they can only select predefined units from the list.

###  Before
image

### After
image

- [ ] Docs updated?


bpedersen2 (approved):

bpedersen2:
As for the E2E failures, see https://github.com/SciCatProject/scicat-backend-next/issues/1013

bpedersen2:
In the longer run we could consider switching the input to a combobox (edit+list) or use  some autosuggest logic. But for now I would say: go ahead and merge

PR #1375, opened 2024-01-18
Swap 3714 scicat in scientificmetadata table edit existing

yoganandaness:
## Description

Validating duplicate name in the Scientific metadata field name. Earlier, this validation was done for the scenario, when a new row of scientific metadata has been added. But for the existing row, if we try to change the field name to an already existing field name, the validation doesn't happen and it allows duplicated values. 

## Motivation

Scientific Metadata validation

## Fixes:

Duplicate validation of the Scientific Metadata field Name. 

## Changes:

Added validation trigger not only for the new Scientific Metadata Row but also for the existing row. Also, optimization has been done to avoid redundant definitions of the Form fields on the Scientific Metadata form. 

## Tests included/Docs Updated? Yes

- [ ] Docs updated?


cypress/e2e/datasets/datasets-metadata.cy.js
```diff
@@ -146,10 +146,88 @@ describe("Datasets", () => {
...
+
+      cy.get("mat-row").contains("Cypress Dataset").first().click();
+
+      cy.wait("@fetch");
+
+      cy.finishedLoading();
+
+      cy.scrollTo("bottom");
+
+      cy.get('[role="tab"]').contains("Edit").click();
```
bpedersen2, line 176: check https://github.com/SciCatProject/frontend/actions/runs/7581087107/job/20648034516?pr=1375#step:7:172

cypress/e2e/datasets/datasets-metadata.cy.js
```diff
@@ -146,10 +146,88 @@ describe("Datasets", () => {
...
+      // cy.get("[data-cy=metadata-name-input]")
+      //   .last()
+      //   .clear()
+      //   .type("name1")
+      //   .parent()
+      //   .click()
+
+      // cy.get("mat-error").contains("Name already exists");
+      // cy.get("button[data-cy=save-changes-button]").should("be.disabled");
+
```
Junjiequan, line 228: Isn't this test the most important part for this PR? Why are we commenting out this

Junjiequan:
## The error suggets 

  1) Datasets
       Add metadata item
         should not be able to edit the existing metadata with a duplicate name:
     CypressError: `cy.type()` failed because the page updated as a result of this command, but you tried to continue the command chain. The subject is no longer attached to the DOM, and Cypress cannot requery the page after commands such as `cy.type()`.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You can typically solve this by breaking up a chain. For example, rewrite:

> `cy.get('button').click().should('have.class', 'active')`

to

> `cy.get('button').as('btn').click()`
> `cy.get('@btn').should('have.class', 'active')`

PR #1388, opened 2024-01-30
Fix: Changed profile.username to profile.displayName for header username display

Junjiequan:
## Description

This PR adapts to the https://github.com/SciCatProject/scicat-backend-next/pull/1041
In the above backend PR we have changed `profile.username` to store `oidc.sub` by default, which in some cases might return string of numbers instead of user's name. So the change of displaying `profile.username` to` profile.displayName` on the header section is required.

- [ ] Docs updated?


src/app/state-management/selectors/user.selectors.ts
```diff
@@ -22,7 +22,7 @@ export const selectCurrentUserName = createSelector(
   selectCurrentUser,
   (profile, user) => {
     if (profile) {
-      return profile.username.replace("ms-ad.", "");
+      return profile.displayName.replace("ms-ad.", "");
```
bpedersen2, line 25: the replace is probably not needed here then.  The backend prependes a marker for the auth source to the username, but  should not do that for the displayName.

bpedersen2 (changes requested):
Good idea, could get some simplification. 

Actually the `replace(' ms-ad',...`) is probably wrong at all places,  as e.g. the `oidc.` prefix would need the same handling. 

Users without a profile should be limited to the system accounts, if I am not mistaken.

src/app/state-management/selectors/user.selectors.ts
bpedersen2: I  checked and all user types do implement the user profile (local, oidc, ldap) . jwt auth relies on the other strategies for initial login/account creation, so we proably can rely on the profile being there (or adopt the BE to ensure that),

nitrosx: @bpedersen2 hopefully I understood your comment above correctly. 
To clarify further: user profile is created for all accounts, no matter which auth strategy they use. It is safe to assume that profile is always present.
I agree with you that the displayName should not be stored in the db with a prefix.
We should display it as it is saved in the database.

nitrosx (changes requested):

nitrosx (approved):

bpedersen2 (approved):

PR #1424, opened 2024-03-12
fix: changed env var naming for create sample button from edit to add

Junjiequan:
## Description

Depends on https://github.com/SciCatProject/scicat-backend-next/pull/1103

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?


bpedersen2:
cypress tests need adoption to changed element name, otherwise looks fine

```
 1) Datasets general
       Create dataset
         should be able to create dataset in the UI:
     AssertionError: Timed out retrying after 10000ms: Expected to find element: `mat-card.add-card`, but never found it.
      at Context.eval (webpack:///../../../cypress/e2e/datasets/datasets-general.cy.js:24:9)
```

Junjiequan:
I will remove Create Dataset and Create sample from the E2E test, because creating dataset and sample form on the UI is incomplete and only supports limited fields. After we improved the form we can then put the test back again.

Junjiequan:
I will remove Create Dataset and Create sample from the E2E test, because creating dataset and sample form on the UI is incomplete and only supports limited fields. After we improved the form we can then put the test back again.

bpedersen2:
> I will remove Create Dataset and Create sample from the E2E test, because creating dataset and sample form on the UI is incomplete and only supports limited fields. After we improved the form we can then put the test back again.

Thats OK

PR #1465, opened 2024-04-27
Configurable facets

Ingvord:
- [x] Proof of the concept
- [x] Refactor dataset-filters.component: extract individual filters
- [x] Fix tests
- [x] Add FilterSettingsDialog test
- [x] Make custom filter UX consistent with dialog
- [x] #1466
- [x] store configuration in ngx state management


nitrosx (commented):
It looks good, although on second thought, I think that the editing and deleting of the individual filter should be in a separate form from the search form. 
As it is now, it can become confusing, as it is not clear if the delete icon is for the content of the field or the filter itself.

nitrosx:
@Ingvord could we create two forms, one to be shown where the filters are now and one to be shown as modal to edit which filters and condition we would like to add?
Please let me know if it makes sense.

nitrosx:
Here my idea based on @Ingvord proof of concept

1. we assume that this user has the filter form empty
configurable_filters_1

2. user wants to add some filters, she clicks the cog wheel on the filter form and an additional modal form appears
configurable_filters_2

3. she now can select which filter she can add to the form. She select the PID
configurable_filters_3

4. and than she select the filter on location
configurable_filters_4

5. now she is done adding filters to the filter form, she clicks apply and the modal closes and she gets the filters as they are currently in the FE
configurable_filters_5

Please notice that removing and filtering of the filters added, can only be done in the settings modal and not in the selection form

nitrosx:
The main idea behind is that the filter form is modified once in a while, so it makes sense to open an additional modal where the user can control the filter form settings and build it as she needs/wants.
We will save the filters form settings in the user settings entry in db, so they can will be used as default until the user decides to change.

Ingvord:
@nitrosx thanks for your idea, explanation, the mockups and putting this all together here! I agree with you. Let's introduce a modal dialog.

despadam (commented):
image

As discussed in the meeting on May 28:
1. It would be more clear to users if the buttons for **Reset** and **Configure** had **both an icon and a corresponding text**.
    - **Reset**: The icon is not very indicative. A suggestion would be to replace it with something like [this](https://fontawesome.com/icons/rotate?f=classic&s=solid), to show that we are starting over.
    - **Configure**: The icon corresponds more to Settings options instead of filter. One icon suggestion to replace it was [this](https://fontawesome.com/icons/filter?f=classic&s=solid). Regarding the button's text, perhaps "Configure" would again be confusing and more relevant to settings, so instead we could use "Add Filters", "Add More", "More" or something similar.

2. After clicking the button to add more filters, a pop-up appears, but I was not able to reproduce locally what was presented during the meeting, so I don't have a screenshot. In that pop-up we had the option to change the order of the additional filters, and each line had on the left-hand side an icon to indicate you can do a drag-n-drop. That icon was the same as in the Reset button (as seen in the screenshot above). We noticed that it can be confusing to use the same icon for different operations, but also normally the sorting operation has an icon where these bars are completely aligned with each other. Hence, the suggestion was to replace it with something like [this](https://fontawesome.com/icons/bars?f=classic&s=solid).

Thank you for the great work @Ingvord!

Ingvord:
@despadam Thanks a lot for your feedback!

> 1. ...

Totally agree 

> 2. ...

The icon indicated usually used for for a menu in many other applications. I guess we can stick with the existing one, as we will change "reset/clear" button icon anyway

Ingvord:
Apparently there are some changes picked up from the master which make the overall UI/UX much more consistent., thanks to @Junjiequan .  I took the liberty to do minor adjustments though:

Peek 2024-06-07 17-21

Please note the icons and labels:

Screenshot_20240607_172705

nitrosx:
@Ingvord it looks good!!!
The only concern that I have is that the two reset buttons are not exactly behaving the same way. I think that for consistency sake, we should make sure that the text explicitly says what they do: _**clear filters**_ and _**clear text**_

One more thing: I checked and the three horizontal lines slightly tilted are indeed the official icon for **_clear all_**.
I will go with what the community decide.

Ingvord:
@nitrosx Thanks for your "in no time" comment :)

Regarding the icons I refer to this resource: https://fonts.google.com/icons

> The only concern that I have is that the two reset buttons are not exactly behaving the same way.

My intention is to make them consistent in a way that they both do *clear* and *apply*. My motivation - they are located in a "major action" section area and have the same look-n-feel as *apply*/*search*, so I would expect them to perform something more than just *clear*. Ah and we lost "x" from full text, which should do only *clear*...

However the functionality requires more love for sure  :cowboy_hat_face:

Ingvord:
All green :champagne:  :clinking_glasses:

src/app/datasets/datasets-filter/filters/pid-filter.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  appConfig = this.appConfigService.getConfig();
+
+  constructor(
+    public appConfigService: AppConfigService,
+    private store: Store,
+  ) {
+    super();
+    this.subscription = this.pidSubject
+      .pipe(
+        skipWhile((terms) => terms.length < 5),
```
Ingvord, line 35: I really think this is redundant now. As we force user to explicitly click the action button, this just introduces a lot of confusion.

Ingvord: Done in [c48ec59](https://github.com/SciCatProject/frontend/pull/1465/commits/c48ec5985f28c9f60e3c7a97fdf5ebbc05c694f8)

Ingvord:
This one is ready for the final review.

@nitrosx I will assign this PR to you. Please feel free to assign to whomever you think it is important to review and test it. Thanks!

bpedersen2 (commented):
Looks  good, I'll try a test build  the next  days.

nitrosx:
@Ingvord Great Job!!! I really like the new filtering form.

Just few items that we can and should discuss later:
- "Text filter" in the form has the same functionality as the "Search bar" at the top. Should we keep it and just disable it or remove it?
- Filters apply to high level fields, conditions apply to scientific metadata. They behave differently. We should brainstorm and invest some resources to uniform the behavior.  
- I would change the title of the form from "Filter by..." to "Filters and Conditions". In my opinion this would be more inline with the previous bullet point

Non blocking feedback. We should have more people testing and getting the feel for it, and brainstorm afterward using the feedback provided from multiple people.

nitrosx (changes requested):
Great job!!!
Last few things to fix or suggests:
- Reset button in Search box: change label to "Reset Search"
- Reset button in Filter box: change label to "Reset Filters"
- Reset filters button should clear values in filters
- Remove Text Filter as it duplicates the search functionalities
- In modal, change "Add Filters" to "Add Conditions"
- Separate list in two sections: filters and conditions. Filters should sort-able, Condition no.
- When adding a new condition, enable it by default
- Add editing to condition in modal

nitrosx (commented):
When I tested the condition edit functionality, I found myself in the situation visible below:
image

nitrosx:
I'm not oppose to merge this PR as it is and create bug issues to be fixed later on.

nitrosx:
@bpedersen2 what is your opinion?

nitrosx (approved):

bpedersen2 (approved):
Looks mostly fine.

Looks fine and merging to master seems fine to get more manual testing.

2 things i see:
1) needs a rebase onto current master due to conflicts
2) It would be nice if the commits could be squashed a bit  (get rid of all those fix lint...) or else do a squash-merge, please.

nitrosx:
@Ingvord please go ahead and do a squash merge

PR #1472, opened 2024-04-30
Fix OIDC login with backend v3.x

### LINKED ISSUE #1467: OIDC authentication is broken with v3 backend
### Summary
OIDC authentication is broken in the frontend when using the backend v3 [see this thread](https://github.com/SciCatProject/scicatlive/pull/144#issuecomment-2082066908)

### Steps to Reproduce

Checkout at `edd3e7041bb297373cf26bab51728a9a2d55aed8` or more recent commit. OIDC login succeeds with v4 but fails with v3

fpotier:
## Description

OIDC login fails in the frontend when using version > v4.2.1 and backend v3.x

## Motivation

Make OIDC login available with both backends in `scicatlive`
If possible, can you tag a new version with this fix so we can use it asap in `scicatlive`

## Fixes:

#1467

### DIFF
--- a/src/app/users/login/login.component.ts
+++ b/src/app/users/login/login.component.ts
@@ -131,8 +131,14 @@ export class LoginComponent implements OnInit, OnDestroy {
       // which are parsed here.
       if (params.returnUrl) {
         // dispatching to the loginOIDCAction passes information to eventually be added to Loopback AccessToken
-        const accessToken = params["access-token"];
-        const userId = params["user-id"];
+        let accessToken = params["access-token"];
+        let userId = params["user-id"];
+        // Required for backend v3 compatibility (access-token and user-id are encoded in returnUrl)
+        if (!accessToken && !userId) {
+          const urlqp = new URLSearchParams(params.returnUrl.split("?")[1]);
+          accessToken = urlqp.get("access-token");
+          userId = urlqp.get("user-id");
+        }
         this.store.dispatch(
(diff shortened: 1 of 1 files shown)


src/app/users/login/login.component.ts
```diff
@@ -131,8 +131,14 @@ export class LoginComponent implements OnInit, OnDestroy {
       // which are parsed here.
       if (params.returnUrl) {
         // dispatching to the loginOIDCAction passes information to eventually be added to Loopback AccessToken
-        const accessToken = params["access-token"];
-        const userId = params["user-id"];
+        let accessToken = params["access-token"];
+        let userId = params["user-id"];
+        // Required for backend v3 compatibility (access-token and user-id are encoded in returnUrl)
+        if (!accessToken && !userId) {
+          const urlqp = new URLSearchParams(params.returnUrl.split("?")[1]);
```
dylanmcreynolds, line 138: For parsing the query portion of a URL, I would consider using the URL class:
Something along the lines of:

```
const query = new URL(params.returnUrl);
accessToken = query.get("access-token");
userId = query.get("user-id");
```

fpotier: `params.returnUrl` is not a valid URL because it only has the path + query parameters. (`message:"URL constructor: /user?access-token=xxx&user-id=662b69f8a283cf65fcfe5191 is not a valid URL."`)

dylanmcreynolds (approved):

PR #1505, opened 2024-06-14
Integrate webix datatable

Ingvord:
Proof of the concept

Peek 2024-06-15 01-57

### DIFF
--- a/src/app/datasets/dataset-datatable/dataset-datatable.component.ts
+++ b/src/app/datasets/dataset-datatable/dataset-datatable.component.ts
@@ -1,0 +1,232 @@
+import {
+  Component,
+  OnDestroy,
+  OnInit,
+  ElementRef,
+  Input,
+  Output,
+  EventEmitter,
+  OnChanges,
+  SimpleChanges,
+  ViewChild,
+} from "@angular/core";
+import * as webix from "webix/webix.js";
+import { Store } from "@ngrx/store";
+import { Subscription } from "rxjs";
+import {
+  clearSelectionAction,
+  selectDatasetAction,
+  deselectDatasetAction,
+  selectAllDatasetsAction,
+  changePageAction,
+  sortByColumnAction,
+} from "state-management/actions/datasets.actions";
+import {
+  selectDatasets,
+  selectDatasetsPerPage,
+  selectPage,
+  selectTotalSets,
+  selectDatasetsInBatch,
+} from "state-management/selectors/datasets.selectors";
+import { Dataset, TableColumn } from "state-management/models";
+import { AppConfigService } from "app-config.service";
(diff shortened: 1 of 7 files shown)


src/app/datasets/dataset-datatable/dataset-datatable.component.ts
```diff
@@ -0,0 +1,232 @@
...
+              this.getTopParentView().refreshColumns();
+            }
+          },
+        },
+      },
+    };
+    this.webixUi.refreshColumns();
+  }
+
+  getWebixColumns(): any[] {
```
Ingvord, line 169: This method shows various possible (not all though) column configuration options. See [Configuring Columns](https://docs.webix.com/datatable__columns_configuration.html) for more details

src/app/datasets/dataset-datatable/dataset-datatable.component.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  ngOnDestroy(): void {
+    this.subscriptions.forEach((subscription) => subscription.unsubscribe());
+    if (this.webixUi) {
+      this.webixUi.destructor();
+    }
+  }
+
+  initializeWebixTable(): void {
+    this.webixUi = webix.ui({
```
Ingvord, line 95: webix [Datatable](https://docs.webix.com/datatable__index.html) creation and configuration

nitrosx:
Thanks for the proof-of-concept.
I'm hesitant to include another full fledged toolkit in the FE just to add the table feature.
I worry that we are going to complicate more the code and make it even more difficult to manage and maintain.

Ingvord:
Hi Max, thanks for the feedback!

Webix has many more widgets, not just tables. I would assume to replace all
visual components with it, hence provide consistent look and feel
experience.

However,  I agree with you. This will introduce whole new abstraction layer
and may increase complexity and maintenance costs.

sbliven:
We talked about this PR in the meeting today. A couple concerns came up.

1. **Pricing** Are all the features we would need available in webix standard (the GPL version)? IMO webix pro licenses are out of the question.
2. **Client-side filtering**. I think users would be very confused that filters are done client side and apply only to the current page of results. I think it needs to be done server-side like the facets, or else it should be omitted from the table.
3. **Maintenance**. It seems like adding webix would increase the complexity of the frontend significantly, and could increase the maintenance burden. I get that this is just a proof of concept, but while the sorting and resizing features shown here would be nice they would hardly be worth it if they make the frontend harder to maintain and develop. Do you see some "killer features" that would have a high impact on the user experience that would come with webix but not available with the current framework?

Ingvord:
Following the discussions at SciCatCon 2024 I will repurpose this PR to demonstrate how to plugins in SciCat FE

PR #1554, opened 2024-08-14
fixes #1553

### LINKED ISSUE #1553: Nginx as less privileged User
### Summary
In order to run the `frontend` image in a restricted environment,
for example where [pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/_ are set to `restricted`, the 
`USER root` is not allowed. 
This also applies for Openshift afaik.

### Steps to Reproduce
try to run the image on a kubernetes cluster where `spec.containers[*].securityContext.runAsNonRoot` is enforces.

### Current Behaviour
If fails

### Expected Behaviour
it runs, root should not be needed in this context

### Extra Details
In general, port 80 is considered a "high privilege" port in Linux. 
It is sufficiemt for the reverse proxy to run with this (and port 443) port.
The upstream server does not need to run on this port.

belfhi:
as described in the issue, the user does not need to be `root`, even for _nginx_.
This PR sets the `nginx.conf` the way the official [docs](https://hub.docker.com/_/nginx) describe what is needed to run as a rootless user.

### DIFF
--- a/scripts/nginx.conf
+++ b/scripts/nginx.conf
@@ -2,10 +2,8 @@
 #   * Official English Documentation: http://nginx.org/en/docs/
 #   * Official Russian Documentation: http://nginx.org/ru/docs/
 
-user nginx;
 worker_processes auto;
-error_log /var/log/nginx/error.log;
-pid /run/nginx.pid;
+pid /tmp/nginx.pid;
 
 # Load dynamic modules. See /usr/share/doc/nginx/README.dynamic.
 include /usr/share/nginx/modules/*.conf;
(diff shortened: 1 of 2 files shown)


scripts/nginx.conf
```diff
@@ -18,9 +16,13 @@ http {
     log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                       '$status $body_bytes_sent "$http_referer" '
                       '"$http_user_agent" "$http_x_forwarded_for"';
-
-    access_log  /var/log/nginx/access.log  main;
-
+                      
+    client_body_temp_path /tmp/client_temp;
```
bpedersen2, line 20: I would prefer to keep the original locations ( var/cache/nginx/... ) and instead fix permissions  in the image build.

bpedersen2 (changes requested):
Generally a good idea.

But as it breaks existing installations, it needs prominent mentions in changelog and readme, and probabaly an announcement in the scicat mailinglist at least.

scripts/nginx.conf
belfhi: That's also an option. It was more along the line of "just log to stdout / stderr if it's running in a container".

bpedersen2: See also #1657

Junjiequan:
Ths issue is addressed in this PR #1657

PR #1563, opened 2024-08-26
Load/Upload user filters and conditions from/to be

Ingvord:
- [x] Load default user settings from the server 
- [x] Load/Upload current user filters and conditions from/to BE

Resolves https://github.com/SciCatProject/scicat-backend-next/issues/616


src/app/shared/services/user-settings.service.ts
```diff
@@ -0,0 +1,21 @@
...
+    "/" +
+    LoopBackConfig.getApiVersion() +
+    "/Users/settings/default";
+
+  constructor(private http: HttpClient) {}
+
+  getDefaultSettings(): Observable<UserSettingInterface> {
+    return this.http.get<UserSettingInterface>(this.defaultSettingsUrl);
+  }
+}
```
Junjiequan, line 21: Please move this to `/shared/sdk/services/custom/User.ts`

Junjiequan (changes requested):

src/app/shared/services/user-settings.service.ts
Ingvord: Why?

Junjiequan: Are you asking why we shouldn't have specific class for this endpint or?

Ingvord: No, sorry, I am asking why you want to move it. It seems the destination is auto-generated.

Junjiequan: The SDK was generated long ago but no longer maintained. 
We directly modify the corresponding files in the `SDK/services/custom` until we replace it with new SDK.

Ingvord: Sounds scary :fearful:

Junjiequan: Yes, it is bad. we need to replace the old one with new SDK asap.

Ingvord: Well, not only we need to upgrade it asap, but I would strongly recommend NOT to keep auto-generated code in the repo. Way better is to include it via dependency, I believe there is some progress already in that direction, right?! Also this will help to clear which API version is being used

Junjiequan: I completely agree with your point. I’ll make sure to pass this feedback along when I next catch up with the colleague who managed this, although they left quite a few years ago. :) It’s definitely something we should keep in mind as we move forward with upgrading and managing dependencies.

src/app/shared/sdk/models/Dataset.ts
```diff
@@ -44,6 +44,9 @@ export interface DatasetInterface {
   createdAt?: Date;
   updatedAt?: Date;
   instrumentId?: string;
+  proposalIds?: string | string[];
+  sampleIds?: string | string[];
+  instrumentIds?: string | string[];
```
nitrosx, line 49: This should not be necessary. 
Datasets output DTO should be the same as before

src/app/datasets/dataset-detail/dataset-detail.component.html
nitrosx, line 1: The field's names should be singular as the dataset output dto did not change

src/app/datasets/dataset-details-dashboard/dataset-details-dashboard.component.ts
nitrosx, line 1: The field's names should be singular as the dataset output dto did not change

nitrosx (changes requested):

nitrosx (approved):

PR #1569, opened 2024-08-27
Fix #1568

### LINKED ISSUE #1568: Search functionality fails after filter reset.
## Issue Name
Search functionality fails after filter reset.

### Summary
After clicking the "Reset Filters" button, the search function no longer works with the same PID value.

https://github.com/user-attachments/assets/27d4de80-49b6-4d5c-9a29-94cf94c457bc

### Steps to Reproduce

1. Apply a search using a specific PID value.
2. Click the "Reset Filters" button.
3. Attempt to search again using the same PID value.

### Current Behaviour
The search does not return any query results after filters are reset.

### Expected Behaviour
The search should return query results when using the same PID value after resetting the filters.

### Extra Details
Provide all the relevant details related to this issues. Include screenshot, wireframes and mockups if helpful.

__PS__: make ssure to add the proper tags to the issue

Ingvord:
subj

### DIFF
--- a/src/app/shared/modules/filters/pid-filter.component.ts
+++ b/src/app/shared/modules/filters/pid-filter.component.ts
@@ -30,7 +30,7 @@ export class PidFilterComponent
   ) {
     super();
     this.subscription = this.pidSubject
-      .pipe(debounceTime(500), distinctUntilChanged())
+      .pipe(debounceTime(500))
       .subscribe((pid) => {
         const condition = !pid ? "" : this.buildPidTermsCondition(pid);
         this.store.dispatch(setPidTermsFilterAction({ pid: condition }));


Junjiequan (commented):
Do we still need debounce here? 
I suppose it was needed before to reduce API calls, but now we trigger API call by clicking apply

src/app/shared/modules/filters/pid-filter.component.ts
```diff
@@ -30,7 +30,7 @@ export class PidFilterComponent
   ) {
     super();
     this.subscription = this.pidSubject
-      .pipe(debounceTime(500), distinctUntilChanged())
+      .pipe(debounceTime(500))
```
Junjiequan, line 33: Do we still need debounce here?
I suppose it was needed before to reduce API calls, but now we trigger API call by clicking apply

Ingvord: Good question! I was thinking about it yesterday and I believe it will spam actions otherwise, so keeping it won't harm

Junjiequan: Should be fine for now. 
It can be removed when we support press enter to execute search

Junjiequan (approved):

Issue #1140, opened 2023-07-24
Jobs migration: tests


nitrosx:
### Summary
Port tests from v3.x if any exists and add relevant cases as needed

sbliven:
Is this relating to jobs? Are there any frontend functions relating to jobs? Backend testing is covered under https://github.com/SciCatProject/scicat-backend-next/issues/626.

(Sorry if I unassigned anyone improperly; I first thought this was the backend issue)

Issue #1139, opened 2023-07-24
Jobs migration: job details


nitrosx:
### Summary
Refactor job details and verify that it works with matching release of backend

Issue #1138, opened 2023-07-24
Jobs migration: job list


nitrosx:
### Summary
Refactor job listing and verify that works with matching version of backend

PR #1585, opened 2024-09-12
Update frontend to match with release-jobs backend

despadam:
## Motivation

Make a first try to see whether the MVP `release-jobs` backend is compatible with the frontend.

## Description

To make the frontend compatible with the new Jobs schema from the `release-jobs` backend, we have to update the job list page, the job detail page, the schema in the frontend and all its references to it, as well as the `.spec` test files.

## Fixes:

- https://github.com/SciCatProject/frontend/issues/1138
- https://github.com/SciCatProject/frontend/issues/1139
- https://github.com/SciCatProject/frontend/issues/1140

## Changes:

- Provide new job schema in `src/app/shared/sdk/models/Job.ts` by updating the definition of `Job` class 
- Update the relevant fields whenever an instance of the `Job` class is created:
  - provide `datasetIds` inside `jobParams`
  - rename `emailJobInitiator` to `createdBy` and its value should be `username` instead of `email`
  - rename `creationTime` to `createdAt`
- Update `jobs-dashboard-new.component` and `jobs-detail.component`, so show in the job list and job details pages the new fields of the Job class
- `jobs-dashboard.component` had to be updated too to avoid getting errors, but it appears to no longer be used
- Update the relevant fields whenever an instance of the `Job` class is created in one of the test files

## Backend version

`release-jobs`

### DIFF
--- a/src/app/datasets/admin-tab/admin-tab.component.ts
+++ b/src/app/datasets/admin-tab/admin-tab.component.ts
@@ -45,24 +45,11 @@ export class AdminTabComponent implements OnInit, OnDestroy {
         .subscribe((user) => {
           if (user && this.dataset) {
             const job = new Job();
-            job.emailJobInitiator = user.email;
-            job.jobParams = {};
-            job.jobParams["username"] = user.username;
-            job.creationTime = new Date();
+            job.createdBy = user.username;
+            job.DateTime = new Date();
             job.type = "reset";
-            const fileObj: FileObject = {
-              pid: "",
-              files: [],
-            };
-            const fileList: string[] = [];
-            fileObj.pid = this.dataset["pid"];
-            if (this.dataset["datablocks"]) {
-              this.dataset["datablocks"].map((d) => {
-                fileList.push(d["archiveId"]);
-              });
-            }
-            fileObj.files = fileList;
-            job.datasetList = [fileObj];
+            job.jobParams = {};
+            job.jobParams["datasetIds"] = [this.dataset["pid"]];
             console.log(job);
             this.store.dispatch(submitJobAction({ job }));
           }
--- a/src/app/datasets/admin-tab/admin-tab.component.ts
+++ b/src/app/datasets/admin-tab/admin-tab.component.ts
@@ -46,7 +46,7 @@ export class AdminTabComponent implements OnInit, OnDestroy {
           if (user && this.dataset) {
             const job = new Job();
             job.createdBy = user.username;
-            job.DateTime = new Date();
+            job.createdAt = new Date();
             job.type = "reset";
             job.jobParams = {};
             job.jobParams["datasetIds"] = [this.dataset["pid"]];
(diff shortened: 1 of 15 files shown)


src/app/shared/sdk/models/Job.ts
```diff
@@ -3,34 +3,43 @@
 declare var Object: any;
 export interface JobInterface {
   id?: string;
-  emailJobInitiator: string;
-  type: string;
-  creationTime?: Date;
-  executionTime?: Date;
+  ownerUser?: string;
```
sbliven, line 6: Not relevant to this PR, but why didn't we call this `owner` on the backend to match `Dataset.owner`?

src/app/shared/sdk/models/Job.ts
```diff
@@ -3,34 +3,43 @@
...
+  type?: string;
```
sbliven, line 7: Should this be required?

src/app/shared/sdk/models/Job.ts
```diff
@@ -3,34 +3,43 @@
...
+  "contactEmail": string;
+  "configVersion": string;
   "jobResultObject": any;
   "createdBy": string;
   "updatedBy": string;
   "createdAt": Date;
   "updatedAt": Date;
+  "ownerGroup": string;
+  "accessGroups": any;
+  "isPublished": boolean;
```
sbliven, line 41: All Ownables now have a top-level `isPublished` field? That doesn't seem correct (but that's not a problem with this PR).

src/app/datasets/admin-tab/admin-tab.component.ts
```diff
@@ -45,24 +45,11 @@ export class AdminTabComponent implements OnInit, OnDestroy {
...
-              files: [],
-            };
-            const fileList: string[] = [];
-            fileObj.pid = this.dataset["pid"];
-            if (this.dataset["datablocks"]) {
-              this.dataset["datablocks"].map((d) => {
-                fileList.push(d["archiveId"]);
-              });
-            }
-            fileObj.files = fileList;
```
sbliven, line 64: We should discuss this change with some site that used the `files` property.

src/app/datasets/archiving.service.ts
```diff
@@ -37,12 +37,8 @@ export class ArchivingService {
 
     const data = {
       jobParams,
-      emailJobInitiator: user.email,
```
sbliven, line 40: Doesn't `contactEmail` still need to be set to `user.email`?

src/app/datasets/datafiles-actions/datafiles-action.component.spec.ts
```diff
@@ -177,7 +177,7 @@ describe("1000: DatafilesActionComponent", () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(DatafilesActionComponent);
     component = fixture.componentInstance;
-    component.files = structuredClone(actionFiles);
```
sbliven, line 180: Weird this wasn't defined before. Your change looks fine for a unit test though.

sbliven (approved):
Up to now I think the backend was backwards-compatible with v3. We should clearly indicate in the README that it requires scicat-backend-next.

PR #1625, opened 2024-10-25
fix(proposal_logbook): inifinite logbook fech loop after RxJS upgrade…

Junjiequan:
… to v7.4.0

## Description
The issue likely arises because, in RxJS 7, observables are more “strict” in how they handle subscriptions. In RxJS 6, subscribing directly to the observable would not necessarily re-trigger the observable emissions if the component re-rendered, as long as the underlying state hadn’t changed. However, in RxJS 7, observables are often “hotter” in behavior, meaning they emit values more readily, even on re-renders or re-subscriptions.

Without take(1), the observable keeps emitting whenever the component or state updates, re-triggering the fetchLogbookAction call. By using take(1), you ensure that the observable completes after emitting a single value, preventing repeated subscriptions and, consequently, avoiding the infinite fetch loop.

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Bug Fixes:
- Fix infinite fetch loop in proposal logbook component by ensuring the observable completes after emitting a single value using take(1).


martin-trajanovski (approved):

martin-trajanovski (commented):
I am happy to approve this but not sure if we should test it somehow so we don't face the same issue in the future.

Junjiequan:
Writing the e2e test for the infinite fetching issue is taking a bit longer than expected for this small but critical fix. 
I will merge this change without the e2e test and open a new PR for it.

PR #1647, opened 2024-11-08
refactor: new sdk frotnend code refactor part 4

martin-trajanovski:
## Description

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

### DIFF
--- a/src/app/datasets/admin-tab/admin-tab.component.html
+++ b/src/app/datasets/admin-tab/admin-tab.component.html
@@ -2,7 +2,7 @@
   <ng-container *ngIf="datablocks$ | async as datablocks">
     <h3>Datablocks: {{ datablocks.length }}</h3>
     <ul *ngFor="let block of datablocks">
-      <li>{{ block.id }}</li>
+      <li>{{ block._id }}</li>
       <ul>
         <li>{{ block.archiveId }}</li>
       </ul>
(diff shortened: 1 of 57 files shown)


src/app/datasets/admin-tab/admin-tab.component.html
```diff
@@ -2,7 +2,7 @@
   <ng-container *ngIf="datablocks$ | async as datablocks">
     <h3>Datablocks: {{ datablocks.length }}</h3>
     <ul *ngFor="let block of datablocks">
-      <li>{{ block.id }}</li>
+      <li>{{ block._id }}</li>
```
nitrosx, line 5: I would use the field `id`, which is consistent with the data  model.

nitrosx (changes requested):

src/app/datasets/admin-tab/admin-tab.component.html
martin-trajanovski: Yes I see this but in the backend we are not exposing the `id` as `@ApiProperty` in the response we only have `_id` defined and I wasn't sure if this was by purpose or we need to fix it.

martin-trajanovski: I have checked that we don't even have the id property in the datablock schema in the backend. Here it is defined: https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datablocks/schemas/datablock.schema.ts#L26. So maybe this is a bigger issue to address on the backend and define it. But we will need a database migration as well to copy the `_id` field to the new `id` as well and expose it.

PR #1657, opened 2024-11-12
Dockerfile: use unprivileged nginx

cfelder:
This allows running this container w/ arbitrary uid support

## Description

## Motivation
running SciCat in an OpenShift environment w/ arbitrary uids

## Fixes:
* nginx unprivileged

## Changes:
* unprivileged port 8080


bpedersen2 (approved):

bpedersen2:
the e2e tests require the nginx.conf file, don't delete it

bpedersen2 (changes requested):
the e2e tests require the nginx.conf file, don't delete it

Dockerfile
```diff
@@ -8,8 +8,6 @@ RUN npm ci
 COPY . /frontend/
 RUN npx ng build
 
-FROM nginx:1.25-alpine
-RUN rm -rf /usr/share/nginx/html/*
```
bpedersen2, line 12: this cleanup line should probably also stay

bpedersen2 (changes requested):
You missed the correct e2e docker compose file:
github uses `  CI/ESS/e2e/docker-compose.e2e.yaml`  

note the ESS in path.

Corresponding changes for the backend tests are probably needed.

bpedersen2:
needs a rebase to get the restructred e2e tests ( the docker composefiel touched is then correct

bpedersen2 (approved):

bpedersen2 (changes requested):
Should be finalized in  of sciat online meeting .

( i added this review to block merging until then, otherwise the code changes are OK for me.)

bpedersen2:
I added a few more reviewers, as this also requires all downstream consumers of the image to adjust the port mapping.

Probably needs updates in the docs and in scicatlive. 

BE e2e tests should work automatically, as the frontend container only exposes one port, so that  traefik will pick it up correctly (https://doc.traefik.io/traefik/providers/docker/#port-detection).

bpedersen2:
> Probably needs updates in the docs and in scicatlive.
 
Scicatlive is also fine as it also uses traefik

Dockerfile
```diff
@@ -8,8 +8,10 @@ RUN npm ci
 COPY . /frontend/
 RUN npx ng build
 
-FROM nginx:1.25-alpine
+FROM nginxinc/nginx-unprivileged
```
dylanmcreynolds, line 11: Should we pin to a version here? `nginxinc/nginx-unprivileged:1.27`?

Dockerfile
```diff
@@ -8,8 +8,10 @@ RUN npm ci
...
+USER root
 RUN rm -rf /usr/share/nginx/html/*
-COPY --from=builder /frontend/dist/ /usr/share/nginx/html/
 COPY scripts/nginx.conf /etc/nginx/nginx.conf
-EXPOSE 80
+USER 101
```
dylanmcreynolds, line 15: Tested in both podman and docker, both give the warning:

`nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:5`

This is because the nginx.conf file we mount in adds `user nginx;`. Is there a particular reason this `USER 101` was added here?

cfelder: The 'cleanup' in line 13 requires switching to ``root``, but the default user is ``nginx (101)``. This uid can be configured in the upstream image using an UID argument at build time.

https://github.com/nginxinc/docker-nginx-unprivileged/blob/0cecd4eba966b7d70fa97619625e22003c93be90/Dockerfile-debian.template#L12

Nevertheless I agree that the user directive in the nginx.conf file does not make sense in this scenario.

Dockerfile
cfelder: I am wondering if that's necessary.

Dockerfile
nitrosx: Probably is better to have a specific version, although who is going to be in charge to keep it up-to-date?

Dockerfile
nitrosx: Do we have a solution to this?

Dockerfile
nitrosx: Is there a solution to this?

nitrosx (commented):
@bpedersen2 can you express your opinion on this?
I'm not sure I am knowledgeable enough to provide an unbiased review

Dockerfile
cfelder: The current stable release is:

docker.io/nginxinc/nginx-unprivileged:1.26.3

Dockerfile
cfelder: Removed user directive in nginx.conf

Dockerfile
nitrosx: I assume that this issue is solved.

nitrosx (approved):

nitrosx:
@bpedersen2 could you please review and approved, if possible?

cfelder:
Apparently one test is failing after merging master into this PR. Unfortunately, I cannot judge if this is a glitch or a reasonable failure. I would really appreciate if we can move this PR forward to get rid of manually patching our own fork for an unprivileged deployment. ptal @bpedersen2.

bpedersen2 (approved):

GBirkel:
https://github.com/SciCatProject/frontend/commit/d7bfbbb95f74548151f40683e47d0ac5ddb8814f#r162312535

This change breaks SciCat Live's config.json-merging feature in merge.json.sh, specifically the line apk update && apk add jq gettext, because apk does not exist on this image.

https://github.com/SciCatProject/scicatlive/blob/57731e4bf81dd902f395dbe47a24a9039e2576cd/entrypoints/merge_json.sh#L5

https://github.com/SciCatProject/scicatlive/blob/57731e4bf81dd902f395dbe47a24a9039e2576cd/services/frontend/compose.base.yaml#L12

SciCat live's front end can no longer find its back end, partly because it doesn't know the URL, and partly because the proxy is not forwarding to the correct port (needs to be 8080 now).

PR #1673, opened 2024-11-27
fix: optimize condition editing logic in DatasetsFilterSettingsComponent

Junjiequan:
## Description
This PR addresses the following issues:
1. Fixes a UI issue in the filter where editing a scientific metadata condition caused the previous condition to disappear.
2. Fixes duplicate creation of scientific metadata conditions
3. Resolves an issue where the input value changes only when the input field is focused again.

image
image
image

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Optimize the condition editing logic in the DatasetsFilterSettingsComponent to fix UI issues related to scientific metadata conditions and input value changes.

Bug Fixes:
- Fix a UI issue where editing a scientific metadata condition caused the previous condition to disappear.
- Resolve an issue where the input value changes only when the input field is focused again.

### DIFF
--- a/src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
+++ b/src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
@@ -66,6 +66,7 @@ export class DatasetsFilterSettingsComponent {
           if (existingConditionIndex !== -1) {
             this.snackBar.open("Condition already exists", "Close", {
               duration: 2000,
+              panelClass: ["snackbar-warning"],
             });
             return;
           }
--- a/src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
+++ b/src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
@@ -23,6 +23,7 @@ import {
   FilterConfig,
 } from "../../../shared/modules/filters/filters.module";
 import { isEqual } from "lodash-es";
+import { MatSnackBar } from "@angular/material/snack-bar";
 
 @Component({
   selector: "app-type-datasets-filter-settings",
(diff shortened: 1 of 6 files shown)


src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
```diff
@@ -56,6 +56,15 @@ export class DatasetsFilterSettingsComponent {
       .subscribe((res) => {
         if (res) {
           const { data } = res;
+
+          // If the condition already exists, do nothing
+          const existingConditionIndex = this.data.conditionConfigs.findIndex(
+            (config) => isEqual(config.condition, data),
+          );
+          if (existingConditionIndex !== -1) {
+            console.log("Condition already exists");
```
martin-trajanovski, line 65: Do we need to console.log here or we just want to return if the condition is met?

martin-trajanovski (commented):
Seems fine to me. Maybe we can write a short e2e test that covers this bugfix or if you think it is not so critical we can cover it in some other scenario later.

src/app/datasets/datasets-filter/settings/datasets-filter-settings.component.ts
Junjiequan: Thanks for pointing that out. 
Ideally, we should show a snack bar popup with the msg to let the user know what's happening, but I've noticed that we have disabled the snack bar on staging and production.
I've kept the waning in a console.log, in case we want to improve the snack bar popup in the future.

nitrosx (approved):

PR #1701, opened 2025-01-13
fix: sdk new package name and some small improvements

martin-trajanovski:
## Description
1. Use the new @scicatproject/scicat-sdk-ts-angular package
2. fix the sdk local generation after introducing some new parameters on the backend

## Motivation
The sdk package was renamed to get better separation for different usage

## Changes:

* changes are mostly renaming the package from @scicatproject/scicat-sdk-ts to @scicatproject/scicat-sdk-ts-angular
* and some other minor fixes

## Tests included

## Documentation

## Backend version

### DIFF
--- a/src/app/state-management/effects/datasets.effects.spec.ts
+++ b/src/app/state-management/effects/datasets.effects.spec.ts
@@ -92,6 +92,7 @@ describe("DatasetEffects", () => {
             "datasetsControllerCreateAttachment",
             "datasetsControllerFindOneAttachmentAndUpdate",
             "datasetsControllerFindOneAttachmentAndRemove",
+            "datasetsControllerAppendToArrayField",
             "datasetsControllerCount",
           ]),
         },


src/app/datasets/batch-view/batch-view.component.ts
```diff
@@ -116,45 +119,75 @@ export class BatchViewComponent implements OnInit, OnDestroy {
...
+              (!this.isAdmin &&
+                !this.userProfile.accessGroups.includes(dataset.ownerGroup))
+            ) {
+              continue;
+            }
+
+            try {
+              await firstValueFrom(
+                this.datasetService.datasetsControllerFindByIdAndUpdate(
+                  dataset.pid,
```
Junjiequan, line 142: I'm not sure about the continue statement there.
Previously the if condition checks: 
1.` if user is not admin and not the owner` do nothing.
2. `if user is not admin and not part of the accessGroups` do nothing.
but now it became continue with try block if above conditions are true.

Maybe it's better of to make a stash and create a new PR to address this refactoring

Junjiequan (changes requested):

Junjiequan (approved):

PR #1933, opened 2025-07-08
(feat) add help cards in /help and /about, based on configurable supportEmail

omkar-ethz:
## Description
This PR adds an environment variable `supportEmail`. If set, a card "Questions or issues" is added in the `/help` page like so:
Screenshot from 2025-07-08 14-55-16

Moreover, only if facility==PSI, an additional "Getting help" card is also added in `/about`:
Screenshot from 2025-07-08 14-56-42

Here, the second line with email is only shown if `supportEmail` is set in config, otherwise only the first line is shown.

## Motivation
Currently, users had no way to look up a contact email in case of issues or questions. This PR adds it via a configurable `supportEmail` env var. (The "Getting help" card in /about is only enabled for PSI, but can be for other facilities as well if desired)

## Fixes:

None

## Changes:

* add `supportEmail` in appConfig
* add a card in /help component
* add a card in /about component

## Tests included

## Documentation

N/A

## Backend version


omkar-ethz:
> what's the idea behind having the email in both places? We should remember to document this once [this](https://www.scicatproject.org/documentation/Development/v3.x/Configuration.html#scicat-frontend) (reviewed and improved) is in this repo's readme/docs

this was based on Leo's suggestion (that user's might also want to see the support email in about). But i agree the Help page is pretty easy to find, so fine with removing the card from About.

omkar-ethz:
> > what's the idea behind having the email in both places? We should remember to document this once [this](https://www.scicatproject.org/documentation/Development/v3.x/Configuration.html#scicat-frontend) (reviewed and improved) is in this repo's readme/docs
> 
> this was based on Leo's suggestion (that user's might also want to see the support email in about). But i agree the Help page is pretty easy to find, so fine with removing the card from About.

I will also create a PR in the current version of the docs once we decide the behavior of `supportEmail` (whether we keep the card in About or not) - so that when the docs are moved to this repo, this gets included as well.

PR #1955, opened 2025-08-04
fix: bug when user is logged in and page is refreshed the user should not be logged out if token is valid and not expired

martin-trajanovski:
## Motivation

https://github.com/user-attachments/assets/92e6b74c-03f5-4265-9ef0-97d39f00d9bf

## Summary by Sourcery

Maintain user session across page reload for valid tokens and enforce logout on expired tokens, while refactoring token creation timestamp to a string and updating related tests

New Features:
- Add Cypress E2E test to verify user stays logged in after page reload with valid token
- Add Cypress E2E test to verify user is logged out after reload when token has expired

Bug Fixes:
- Prevent valid authentication tokens from logging out users on page refresh

Enhancements:
- Refactor token handling to store the creation timestamp as a string instead of a Date
- Clear API config accessToken and bearer credentials when the token is expired

Tests:
- Update unit tests for user reducer and selectors to expect the created field as a string
- Adjust initial user store state to use string for token creation timestamp

### DIFF
--- a/src/app/state-management/effects/user.effects.spec.ts
+++ b/src/app/state-management/effects/user.effects.spec.ts
@@ -465,7 +465,7 @@ describe("UserEffects", () => {
         id: "testId",
         ttl: 100,
         scopes: ["string"],
-        created: new Date(),
+        created: new Date().toISOString(),
         userId: "testId",
         user: "testUser",
         rememberMe: false,
@@ -497,7 +497,7 @@ describe("UserEffects", () => {
         id: "testId",
         ttl: 3600,
         scopes: ["string"],
-        created: new Date(),
+        created: new Date().toISOString(),
         userId: "testId",
         user: "testUser",
       } as unknown as SDKToken;
@@ -522,7 +522,7 @@ describe("UserEffects", () => {
         id: "testId",
         ttl: 100,
         scopes: ["string"],
-        created: new Date(1611438651),
+        created: new Date(1611438651).toISOString(),
         userId: "testId",
         user: "testUser",
       } as unknown as SDKToken;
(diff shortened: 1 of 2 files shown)


Junjiequan (commented):
Hi Martin, I tested your fix, but the OIDC login issue still persists.

The problem is that [auth.service.ts > persist()](https://github.com/SciCatProject/frontend/blob/master/src/app/shared/services/auth/auth.service.ts#L52) stringifies a Date Object, resulting in double quotes like:
```
"\"Mon Aug 04 2025 15:30:38 GMT+0200\""
```
Then in this line [auth.service.ts:44](https://github.com/SciCatProject/frontend/blob/master/src/app/shared/services/auth/auth.service.ts#L43) the value returns null since itsan  invalid date.

Followed by `Math.floor(new Date(created).getTime())` in [user.effect.ts:272](https://github.com/SciCatProject/frontend/blob/master/src/app/state-management/effects/user.effects.ts#L273), which returns 0 because `new Date(null).getTime()` equals 0, that results in `isTokenExpired` is always **true**, since currentTime > 0, [this.authService.clear()](https://github.com/SciCatProject/frontend/blob/master/src/app/state-management/effects/user.effects.ts#L277) is triggered everytime.

The root cause is that the persist() function does not handle Date object correctly in the [persist function ](https://github.com/SciCatProject/frontend/blob/master/src/app/shared/services/auth/auth.service.ts#L60) where we need to use `.ISOstring()` to handle Date type properly instead of `stringify`. 

My solution to this would be:
```
      this.storage.set(
        `${this.prefix}${prop}`,
        value instanceof Date
          ? value.toISOString()
          : typeof value === "object"
            ? JSON.stringify(value)
            : value,
        this.token.rememberMe ? expires : null,
      );
```

let me know if it makes sense

PR #1976, opened 2025-08-20
Fixing run number in dataset list

nitrosx:
## Description
This PR removes the custom function to extract the run number from the scientific metadata, but uses the high level field.

## Motivation
We need to show run number high level field by default. A soon to be created PR, will allow to include scientific metadata in the datasets list, so a different run number from scientific metadata will be addressed by the new PR.

## Changes:
* datasets table

## Tests included

## Documentation

## Backend version

### DIFF
--- a/src/app/datasets/dataset-table/dataset-table.component.ts
+++ b/src/app/datasets/dataset-table/dataset-table.component.ts
@@ -213,7 +213,7 @@ export class DatasetTableComponent implements OnInit, OnDestroy {
     this.columns = currentColumnSetting;
     const translated$ = forkJoin(
       currentColumnSetting.map((i) =>
-        this.translateService.get(i.name).pipe(
+        this.translateService.get(i.header || i.name).pipe(
           map((translated) => ({
             ...i,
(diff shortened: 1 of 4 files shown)


Junjiequan (approved):

src/app/datasets/dataset-table/dataset-table.component.ts
Junjiequan, line 378: Please cleanup all the remaining`scientificMetadata.runNumber.value`, if it's not going to be used anymore

Junjiequan (changes requested):

src/app/datasets/dataset-table/dataset-table.component.ts
```diff
@@ -368,7 +372,6 @@ export class DatasetTableComponent implements OnInit, OnDestroy {
   onSortChange(event: SortChangeEvent): void {
     const { active, direction } = event;
     let column = active.split("_")[1];
```
emigun, line 374: Linter says `const colum` 👍

PR #2007, opened 2025-09-16
feat: add ability to merge config files

minottic:
## Description
This is handy when different envs require shared bits from the same config but need overriding some other parts, e.g. the backendUrl. 

## Changes:

* enables merging sequentially up to 10 files, config.json -> ... -> config.3.json -> config.override.json -> ... -> config.override.3.json
* adds config.override* to gitignore to allow scicatlive not to mess with git

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Enable AppConfigService to merge multiple local configuration files and fallback to them when backend config is unavailable, update .gitignore, and add tests for the new merging logic.

New Features:
- Support merging of up to 10 sequential config files (normal and override) in AppConfigService

Enhancements:
- Implement loadAndMerge and mergeConfig methods to sequentially combine multiple local config JSON files
- Refactor loadAppConfig to use the merged local config on backend failure

Tests:
- Add unit tests for mergeConfig to verify multi-file merging behavior
- Add tests to validate loadAppConfig fallback to merged local config and correct HTTP requests

Chores:
- Ignore config.override* files in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ Thumbs.db
 
 # karma
 karma-junit.xml
+
+# config files overrides
+config.override.json
+config.override.*.json
--- a/src/app/app-config.service.spec.ts
+++ b/src/app/app-config.service.spec.ts
@@ -5,7 +5,7 @@ import {
   AppConfigService,
   HelpMessages,
 } from "app-config.service";
-import { of } from "rxjs";
+import { Observable, of } from "rxjs";
 import { MockHttp } from "shared/MockStubs";
 
 const appConfig: AppConfigInterface = {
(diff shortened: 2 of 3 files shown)


HayenNico (approved):
Looks good to me - I assume the main use case for this change is SciCat Live?

For documentation purposes: What is the order of precedence of the different config files?

minottic:
thanks. This is also handy, in our case, as we have very similar configs in DEV and PROD. Being able not to replicate the whole config in PROD but just supplying an additional file with the changes is often handy and reduces copy and paste errors. 

The .override ones, on the contrary, are more for scicatlive as you say, essentially not to mess with git history there.

About documenting, that's a very good idea I think, I wasn't sure where to put it though. 

Do you know where I should add that info?

minottic:
the order is this: 
```
merges sequentially up to 10 files, config.json -> ... -> config.3.json -> config.override.json -> ... -> config.override.3.json

```

PR #2013, opened 2025-09-19
RELEASE:2025-09-10

minottic:
Part of this relaease is to use the unprivileged nginx image, which contains breaking changes (e.g. the port). @Junjiequan plz let me know if you are comfortable releasing this.Thanks!


minottic:
> If I remember correctly, after the release, the port needs to be changed from 80:8080?

yes, I also remember this. 

Thanks! I think it's indeed a new major

Issue #1912, opened 2025-06-26
Change "create dataset" button


sofyalaski:
## Expected Behaviour
Given the current form: 

1.  Users see are the entries from the dataset schema that are common across raw/derived dataset and we enforce users to provide the mandatory fields. ( here user specifies if it's a raw or derived)
2.  Check differences in API V3 and V4 in the DTOs of raw and derived datasets, add compatibility when necessary. Additional fields appear in the questionnaire if any those are needed for any dataset type. 
3. Ask user if they want to provide the schema for the scientific metadata. Must be a reachable URL. 
4. Form expands with the schema fields.
5. Upon save a POST request is sent to the backend
6. Message "Dataset creation successful/failed" and redirect to it

## Current Behaviour
![Image](https://github.com/user-attachments/assets/d0f63028-3928-4993-aac1-0e39d9dafab8)

## Details

There are similar forms in development by Martin on Published Data and from David on OpenEM Ingestor.  We should reuse the technology (jsonforms) and agree on the same style

PR #2040, opened 2025-10-03
Ingestor component

sofyalaski:
## Description
This is a big PR that introduces two changes. 
1. This is a big change and it introduces many new components and a dependency on another backend. This, however, can be turned off in the `config.json` file by setting: 
```
  "ingestorComponent": {
    "ingestorEnabled": false,
    }
```    
2. At the SciCatCon 2025, we discussed a new [project to simplify ingestion](https://github.com/orgs/SciCatProject/projects/20/). This PR also introduces a change to how the button "Create Dataset" below the filter side bar at the Dataset page behaves according to #1912 .  ( Option to control it already present in config - `addDatasetEnabled`

## Motivation
At PSI with OpenEM we have been working on a new ingestor backend that will allow data ingestion from sites different from the host of  SciCat Catalog.  This is represented by Point 1. An addition of the Ingestor backend repo into SciCatProject [is planned](https://github.com/orgs/SciCatProject/projects/20/views/1?pane=issue&itemId=117173574&issue=SciCatProject%7Cscicat-backend-next%7C2014) as well. 

## Changes:
For point 2: 
* when user wants to ingest dataset into SciCat from frontend, the dialog opens, where user enters dataset-specific information. 
* User can provide a url to json schema for scientific metadata. We only check if provided JSON is a valid object. 
* If user provided schema, an additional set of questions is created based on that schema and user can specify the scientific metadata details based on it
* Conformation page where user can review entered metadata
* Dataset is being added after the submission of the form

## Tests included

## Documentation

## Backend version


sbliven (commented):
@sofyalaski, can you add some screenshots of the two options to the description? I think that would make it easier for people to review, since the ingestor isn't that easy to run yourself (yet).

We should also start working on user and operator docs for this feature.

I will leave off formally approving this, as I think that should be done by someone outside of OpenEM. It looks good though, and I'm excited to get this merged into main!

sofyalaski:
Here is an example of dataset ingestion from the main "Datasets" view
https://github.com/user-attachments/assets/44cab5a9-3823-404a-8502-ca5fd9990c3a

sofyalaski:
And with the ingestor component connected on the backend. It's not configured properly, so the trsnafer hasn't succeeded

https://github.com/user-attachments/assets/61535a10-3354-4b36-8bd4-e7abb958a70f

nitrosx:
Can we style the ingestion popup more inline with the rest of the application? 
Also I feel that we are using a lot of space and it requires a lot of scrolling.
image

nitrosx:
@sbliven Just couple of comments:
- we should explicitly say also in the FE that the feature is experimental. I am thinking about a banner in the popup.
- regarding the metadata: if I do not provide any schema, I should be able to add any metadata that I want using the editing provided in the default dataset view.

PR #2064, opened 2025-10-15
docs: adding the workflow yaml and the default  from docs-template

LAShemilt:
## Description
Adding a documentation workflow to the frontend repository based on the `docs-template` repository . This provides unified styling for all documentation across the SciCat project space. 

## Motivation
We want to put documentation next to the microservices , instead of having a single documentation repository which is hard to maintain. We want the styling and CI of the docs to be consistent across the SciCat Project. 

## Changes:
* added default `mkdocs.yml` to `.github/mkdocs` path
* added `publish-docs.yml` workflow

## Summary by Sourcery

Set up documentation deployment via GitHub Actions and provide a default mkdocs configuration for unified styling.

CI:
- Add publish-docs.yml workflow to deploy documentation on pushes to main and version tags using the docs-template action

Documentation:
- Add default mkdocs.yml in .github/mkdocs inheriting from the shared docs-template configuration

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: SciCatProject/docs-template/.github/actions/mkdocs-pages@main
+      - uses: SciCatProject/docs-template/.github/actions/mkdocs-pages@v0.0.1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,17 @@
+name: Deploy docs
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - main
```
Junjiequan: frontend is using `master` branch

Issue #2052, opened 2025-10-08
Add selection icon from detailed dataset view


rkweehinzmann:
## Allow selection of dataset from detailed dataset view
From detailed dataset view it would be nice to directly allow the user to add it to the selection/cart and browse to forward and backward. It’s when looking at the details decisions are made for further consideration or not.

__PS__: make ssure to add the proper tags to the issue

emigun:
Nice!

I think we need to hide this "Add to selection" link when the user is not logged in though. Currently a non logged in user can click the link, it is greyed out but nothing happens.

nitrosx:
We should implement this functionality with configurable actions.
We also should come up with the user flow, so we have clear idea of what to implement

PR #2135, opened 2025-12-04
feat: add "add to Cart" Button to dataset detail view

FloKo2669:
## Description

Adds the "add to Cart" Button  to the dataset detail view to add the current dataset to the batch cart, described in issue #2052.

Screenshot from 2025-12-04 14-56-26

When dataset is already in the cart, button is disabled:
Screenshot from 2025-12-04 14-56-39

Consequently there is some space between the breadcrumb and the tabs to accommodate the button (see pictures).
Before: 

Screenshot from 2025-12-04 15-15-16

## Changes:
* added Button to Dataset Detail View
* added new action "addCurrentToBatch" to add the current selected dataset to the batch
* tests

## Tests included

## Summary by Sourcery

Add support for adding the currently viewed dataset to the batch cart from the dataset detail view and integrate this with existing batch state handling.

New Features:
- Add an "Add to Cart" button to the dataset detail view that dispatches adding the current dataset to the batch cart when shopping cart functionality is enabled.

Enhancements:
- Introduce a dedicated addCurrentToBatch action and reducer logic to append the current dataset to the batch only if it is not already present.
- Expose an isInBatch observable in the dataset details dashboard to disable the add-to-cart button when the current dataset is already in the batch.
- Ensure batch-related side effects also react to the new addCurrentToBatch action.

Tests:
- Add unit tests for the new addCurrentToBatch action and its reducer behavior to prevent duplicates in the batch.

### DIFF
--- a/src/app/datasets/dataset-details-dashboard/dataset-details-dashboard.component.html
+++ b/src/app/datasets/dataset-details-dashboard/dataset-details-dashboard.component.html
@@ -13,8 +13,8 @@
       color="primary"
       class="add-batch"
     >
-      <mat-icon>shopping_cart</mat-icon>
-      Add to Cart
+      <mat-icon>my_library_add</mat-icon>
+      Add to Selection
     </button>
   </div>
   <nav mat-tab-nav-bar [tabPanel]="tabPanel">


bpedersen2:
> Small change requested. Also, we should think about remove `Add to selection` and the publish toggle and use Configurable Action instead. Food for thoughts for a following PR

I created #2139 to track this idea.

PR #2169, opened 2026-01-07
refactor(dataset): redesign & relocate data buttons

abdimo101:
## Description
This PR moves the data buttons from top of the table to the filter section.

Before: 
image

After:
 
image

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Relocate and restyle the datasets public/embargoed data toggle from the table actions toolbar into the filter panel and wire it to the datasets filtering store state.

New Features:
- Expose a public vs embargoed data toggle within the datasets filter panel that drives dataset and facet fetching.

Enhancements:
- Align the public/embargoed toggle styling with the filter card layout for better visual integration and responsiveness.
- Simplify the dataset table actions toolbar by removing the public data toggle logic now handled in the filter component.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -15,7 +15,7 @@ jobs:
       ## Extract information about the dependencies being updated by a Dependabot-generated PR
       - name: Dependabot metadata
         id: dependabot-metadata
-        uses: dependabot/fetch-metadata@v2.4.0
+        uses: dependabot/fetch-metadata@v2.5.0
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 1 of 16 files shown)


Junjiequan (changes requested):
I think what @fpotier commented is a valid point. On top of that we don't have embargoed concept implemented yet, I'd suggest to keep the label as it is for now. @nitrosx any thoughts?

abdimo101:
Hi @fpotier, 

I agree with your opinion and the labels have been changed to make it clearer. 

Concerning your second point about `/api/v3/datasets/fullquery` not returning public data the user does not have access to. I have tested it and it seems to be working as intended for me. 

Can you tell me how you set up the tests that led to the conclusion you provided? Do you get the same problem in the frontend or when using Swagger?

PR #2213, opened 2026-02-04
fix: prevent numerical metadata values from being incorrectly formatted as dates

abdimo101:
## Description
This PR fixes a bug where numeric metadata values (like "10") were being incorrectly displayed as dates (like "1970-01-01 01:00") in the metadata view table. The fix ensures that when metadata has a type field set (such as "number" or "quantity"), that type is respected and the value is displayed correctly.

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Bug Fixes:
- Respect the explicit metadata type field when determining if a value is a date so that numeric and other non-date values are no longer misinterpreted and displayed as dates.

### DIFF
--- a/cypress/e2e/datasets/datasets-metadata.cy.js
+++ b/cypress/e2e/datasets/datasets-metadata.cy.js
@@ -769,6 +769,224 @@ describe("Datasets", () => {
         cy.contains(`unitSI: "${metadata.unitSI}"`).should("exist");
       });
     });
+
+    it("should be able to add a metadata entry with type 'number'", () => {
+      const newMetadataName = "Cypress dataset number metadata";
+      const metadata = {
+        value: "2010",
+        name: "number_metadata",
+        human_name: "Number test",
+        type: "number",
+      };
+      cy.createDataset({
+        type: "raw",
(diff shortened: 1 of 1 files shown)


src/app/shared/modules/scientific-metadata/metadata-view/metadata-view.component.ts
```diff
@@ -240,20 +240,16 @@ export class MetadataViewComponent implements OnInit, OnChanges {
 
   isDate(scientificMetadata: ScientificMetadataTableData): boolean {
     // NOTE: If the type is date, we expect the value to be in ISO format.
-    if (scientificMetadata.type === "date") {
-      return true;
+    if (scientificMetadata.type) {
+      return scientificMetadata.type === "date";
```
Junjiequan: I'm not sure about this change, previous code returns only when type is date but with this change it returns when type exist. Is it really a fix? Are you sure that the codes below should not executed when there is type?

Junjiequan (changes requested):

Issue #1728, opened 2025-01-28
Inconsistent Configuration Fetching for SciCat Frontend


Ingvord:
## Issue Name

Inconsistent Configuration Fetching for SciCat Frontend

### Summary

Currently, there is an inconsistency in how the SciCat frontend fetches its configuration. The frontend supports fetching configuration both locally (from a static file) and remotely (from the backend), leading to unnecessary complexity and potential confusion.

### Steps to Reproduce
 
2.  Start the SciCat frontend application.
3.  Observe how the configuration is fetched:      First, it attempts to fetch configuration from the backend at `/backend/api/v3/admin/config`.
      If the backend configuration is unavailable, it falls back to fetching local configuration from `/scicat/assets/config.json`.
3. Analyze the code behavior and handling in case of errors or missing configurations.

### Current Behaviour

The frontend uses a nested try-catch structure to manage configuration fetching, which is an anti-pattern for flow control.

This approach introduces confusion and tightly couples frontend behavior to backend configuration, violating separation of concerns.

```javascript
try {  
  const config = await this.http  
    .get("/backend/api/v3/admin/config")  
    .pipe(timeout(2000))  
    .toPromise();  
  this.appConfig = Object.assign({}, this.appConfig, config);  
} catch (err) {  
  console.log("No config available in backend, trying with local config.");  
  try {  
    const config = await this.http.get("/scicat/assets/config.json").toPromise();  
    this.appConfig = Object.assign({}, this.appConfig, config);  
  } catch (err) {  
    console.error("No config provided.");  
  }  
}  
```

### Expected Behaviour

The backend should not be responsible for providing frontend configurations.

The frontend configuration should be purely local to the frontend to avoid unnecessary dependencies.

Error handling should avoid the use of `try-catch` for flow control and instead use a cleaner and more modular approach.

### Extra Details

Refactor the configuration fetching logic to remove the backend dependency for frontend configuration.

Ensure error handling is clean and adheres to best practices.

If remote configuration is unavoidable, establish a clear distinction between backend and frontend configuration responsibilities.

bpedersen2:
This has been done in this way on purpose:

For a production deployment everything is served on one hostname,
 and the FE can fetch its config from the corresponding backend.  You  can then use one frontend-deployment to serve different backends on different hostnames.

The FE-local version is only suitable for  dev setups without a proxy included, as  there is no way to find where the backend should be. Maybe we should skip this option and just require dev setups with proxy.

[...]

PR #2290, opened 2026-03-26
fix: take frontend config primarily from the backend

sbliven:
It is now recommended to serve frontend configuration settings from the backend. This has long been discussed (#1728) but the introduction of runtime configuration (https://github.com/SciCatProject/backend/pull/2355) makes it more urgent. This PR introduces changes to the frontend configuration reflecting the expectation that in most scenarios only the URL to the backend must be configured in the frontend container.

## Description

The recommended order of configs to load is now (later entries override earlier):
1. app-config.defaults.ts (built into the container)
2. /assets/config.json (fetched during loading; provides the backend url)
3. [backendURL]/api/v3/admin/config (includes runtime configs and frontend.config.json from the backend container)

This is accomplished by adding a new `additionalConfigs` array to the configuration. Any URLs listed in this array will be fetched at runtime and merged over the existing settings. It can also be used to replicate the old `config.override.json` behavior if desired (which has been removed); just add `/assets/config.override.json` to the additionalConfigs list in `config.json`.

## Motivation

## Fixes:

* #1728

## Changes:

Changes requiring deployment changes:
* [BREAKING CHANGE] **`config.override.json` is not loaded by default.**
  * _Motivation:_ This was previously used to avoid modifying config.json, a common source of git conflicts. This is not a problem with the new minimal config.json, so we can avoid the extra http request.
  * _Quick fix:_ Add `/assets/config.override.json` to the additionalConfigs list in `config.json`.
  * _Better fix:_ Migrate any modified entries to the backend's frontend.config.json file
* config.json defaults to looking for the backend config on the current domain. If this is not the case, include the domain name in the additionalConfigs URL. For instance, for scicatlive:
   ```json
   {
     "additionalConfigs": ["http://backend.localhost/api/v3/admin/config"]
   }
   ```

Other changes:
- Some default values didn't match the type annotations, and have been updated where necessary
- the `allowConfigOverrides` setting has been removed

This PR is a draft pending tests and doc updates.

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Load frontend configuration from built-in defaults and additional runtime config URLs instead of relying on backend admin config and config.override.json.

New Features:
- Introduce a default frontend configuration object embedded in the application image.
- Add support for chaining and merging additional configuration files via an additionalConfigs list, with protection against circular references.

Bug Fixes:
- Align configuration default values and types with their declared annotations.

Enhancements:
- Simplify and harden app configuration loading logic, removing allowConfigOverrides and the special config.override.json mechanism.
- Extend the app configuration schema with siteIcon and labelMaps options and tighten field customization typing.
- Provide richer, opinionated defaults for UI layout, dataset/proposal lists, actions, and labels through the new default config file.

Deployment:
- Change deployment expectations so that frontend config is primarily provided by backend-hosted configuration endpoints rather than per-environment override files.


minottic (commented):
I like this! It adds way more control on what to load, and would both allow changing the BE config URL and remove the 404 when the BE is not at api/v3/admin/config simply be overriding config.json with an empty additionalConfigs.

If I have to find a drawback (personal opinion, but happy anyway with this change!), it's maybe a bit over engineered, I believe we would have achieved the same by moving the BE config load from being the 1st loaded to the last, having prepended the lbBaseUrl living with the 404 or having refactored `enableAdditionalConfig` to `additionalConfigs` (as the change here), but without the recursion (essentially loading config.json, config.override.json and /v3/config only). The administrator would then be able not to touch the config.json (which had config.override.json in the additionalConfigs) and then have a config.override.json with the BE config in the additionalConfigs

nitrosx:
@sbliven can you give me an update on this PR?
Is still relevant? if so, should we make sure to dedicate some effort in to it?
Thanks

minottic:
any update here @sbliven , @Junjiequan ? This would simplify our config management at PSI quite a bit, especially when needing to add temporary banners with downtimes

bpedersen2:
Just my 2c:
 - case 1:   Frontend and backend share a common base url ( standard production deployment)
              ==> relative URLs should work fine, and defaults should work out of the box
 - case 2:  Fronted uses a different base url ( e.g. an alternative FE with special configs)
           ==> needs a way to get at least the base url  for the BE from a location relative to the FE base url
           ==> the BE served config may needs to be distinct from the default or we need  FE scoped values in the config 
( `{ 'base': <common settings>, 'FE1': <settinings only for FE1>, 'FE2': <settings for fe2>}` if we want a singular endpoint.

Maybe split out the FE bootstrap config  (just BE base url and  full config URL )  to another endpoint  to simplify things?

PR #2305, opened 2026-03-31
feat: add source to saved columns and always show user added columns

emigun:
## Description
Add source field to saved settings. If source == "user" the column will be shown even if not part of default settings.

## Motivation
Ground work to allow user added columns to be shown even if they are not part of default columns. Currently all columns that are not part of default columns are dropped.

## Changes:
New field source in saved settings. Header, path and tooltip are also saved, since those could be missing if there is no matching default column.

## Tests included

## Documentation

Still interested in knowing where to update...?

## Backend version

## Summary by Sourcery

Add support for persisting and restoring user-added dataset table columns, including their metadata, so they remain visible even when not part of the default column configuration.

New Features:
- Persist the source, header, path, and tooltip metadata for dataset table columns in user settings to distinguish default from user-added columns.
- Allow user-added columns to be retained and merged into table configurations even when they do not exist in the default column set.

Enhancements:
- Extend table column and field models to include a source flag indicating whether a column is default or user-defined.
- Propagate the new column metadata through dataset table save/load flows to ensure consistent behavior across the UI and settings.

Tests:
- Add unit tests for dataset table settings persistence and table configuration merging to cover user-added columns and default source behavior.

### DIFF
--- a/src/app/datasets/dataset-table/dataset-table.component.ts
+++ b/src/app/datasets/dataset-table/dataset-table.component.ts
@@ -224,15 +224,20 @@ export class DatasetTableComponent implements OnInit, OnDestroy {
   saveTableSettings(setting: ITableSetting) {
     this.pending = true;
     const columnsSetting = setting.columnSetting.map((column, index) => {
-      const { name, display, width, type, format } = column;
+      const { name, header, display, width, type, format, path, tooltip } =
+        column;
 
       return {
         name,
+        header,
         enabled: !!(display === "visible"),
         order: index,
         width,
+        path,
+        source: column.source ?? "default",
         type,
         format,
+        tooltip,
       };
     });
     this.store.dispatch(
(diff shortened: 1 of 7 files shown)


src/app/datasets/dataset-table/dataset-table.component.ts
```diff
@@ -224,15 +224,20 @@ export class DatasetTableComponent implements OnInit, OnDestroy {
   saveTableSettings(setting: ITableSetting) {
     this.pending = true;
     const columnsSetting = setting.columnSetting.map((column, index) => {
-      const { name, display, width, type, format } = column;
+      const { name, header, display, width, type, format, path, tooltip } =
+        column;
 
       return {
         name,
+        header,
```
emigun, line 232: This should make `header` work also for saved settings. I think we can remove `path` and column type `custom`. We should in that case check if someone relies on that feature.

Junjiequan (approved):
LGTM as a ground work.

I assume that next step is to add `userAdded:true` whenever user adds a new column. And the field is added only when column do not exist in the default column, right?

emigun:
> I assume that next step is to add `userAdded:true` whenever user adds a new column. And the field is added only when column do not exist in the default column, right?

Yes exactly. I think the backend handles the part about making sure there are no duplicates in saved settings, but it has to be verified.

PR #2314, opened 2026-04-08
feat: allow published data to be unauth

minottic:
## Description
This makes published data overview and {id} public,
since the endpoint allow non-auth access restricting to registered states.
It hides the buttons when non auth to avoid confusion

## Summary by Sourcery

Make published data overview and details pages publicly accessible while gating modification actions behind login status.

New Features:
- Allow access to the published data dashboard and individual published data details routes without authentication.

Enhancements:
- Hide published data status/action buttons on the details page when the user is not logged in, showing them only for authenticated users.
- Track login state in the published data details component via the store and update tests to cover login-dependent button visibility and store-driven state.


fpotier (approved):
lgtm!

nitrosx (approved):

PR #2316, opened 2026-04-09
fix: use new sdk

fpotier:
## Description
Update sdk calls for next release:
- Replace `SampleClass` by `OutputSampleDto`
- `filter` query param are passed as objects rather than a json string

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Update frontend to use the new SDK data model for samples and query parameters.

Bug Fixes:
- Adjust dataset and job API calls to pass filter parameters as objects instead of JSON strings to match the updated SDK interface.

Enhancements:
- Replace usages of SampleClass with OutputSampleDto across state, components, and tests to align with the new SDK contract.

Tests:
- Update unit tests and shared mocks to use OutputSampleDto and the new API call signatures.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "@ngrx/store": "^19.1.0",
         "@ngx-translate/core": "^17.0.0",
         "@ngxmc/datetime-picker": "^19.3.1",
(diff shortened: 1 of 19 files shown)


minottic (approved):
this will require the new sdk version, right?

PR #2318, opened 2026-04-10
feat: add deleted view when using jobs

minottic:
## Description
It adds a deleted view fetching from "datasetLifecycle.archiveStatusMessage": "deleted", when jobs are enabled

image

It also improves the breadcrumb removing code duplication and reusing the RxJS reducer as well as avoiding to override the skip

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Add a deleted archive view mode for datasets and reuse centralized filter state when navigating via breadcrumbs and archive mode controls.

New Features:
- Expose a new deleted archive view mode in the dataset archive mode selector and dataset list status labels.

Enhancements:
- Restore dataset filters and archive view mode via store actions when navigating from the datasets breadcrumb instead of recomputing filters locally.
- Preserve the existing pagination skip value when changing archive view mode while still applying the appropriate mode filter.
- Allow partial restoration of dataset filters through the setFiltersAction to reuse stored filter state across navigations.

Tests:
- Extend reducer, breadcrumb component, and dataset table actions tests to cover the deleted view mode, filter restoration, and mode-change behavior.

### DIFF
--- a/src/app/datasets/dataset-table-actions/dataset-table-actions.component.spec.ts
+++ b/src/app/datasets/dataset-table-actions/dataset-table-actions.component.spec.ts
@@ -105,6 +105,18 @@ describe("DatasetTableActionsComponent", () => {
         setArchiveViewModeAction({ modeToggle }),
       );
     });
+
+    it("should dispatch selected non-default archive mode", () => {
+      dispatchSpy = spyOn(store, "dispatch");
+      const modeToggle = ArchViewMode.deleted;
+
+      component.onModeChange(modeToggle);
+
+      expect(dispatchSpy).toHaveBeenCalledTimes(1);
+      expect(dispatchSpy).toHaveBeenCalledWith(
+        setArchiveViewModeAction({ modeToggle }),
+      );
+    });
   });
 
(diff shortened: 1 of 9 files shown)


nitrosx (approved):
It looks good to me.
I have two concerns:
- the archive workflow is hard coded
- the dataset lifecycle management

but we can discuss and brainstorm about them at a later stage

minottic:
> the archive workflow is hard coded

I have been playing a bit around with configurable-actions (see #2350 and #2351) and there I replaced the hardcoded actions in batch and the dataset ones with a config. Doing that, I also thought we could replace these tiles with another configurable action, but I was debating since the config would become massive and we would loose all the ngrx store capabilites (e.g. the breadcrumb would break). But probably something to discuss further. Maybe another small comment is that this PR adds to something already hardcoded, so if we want to remove the hardcoded part maybe it should be at a later stage (essentially I agree with your comment). Maybe an easy win would be to provide via the config the list of datasetlifecycle values we are displaying in the tiles and that would be mirrored in the GET filter (to keep the config slim)

> the dataset lifecycle management

what do you mean here?

PR #2321, opened 2026-04-13
feat: configurable link actions

abdellah257:
## Description
Add the option to configure the url of `link` type actions

## Motivation
Support dynamic links that are selection dependent
example:
```json

    "url": "https://<host_name>/proposal/getfile?download=1&file={{ @sourceFolder }}/{{ @file }}",
    "type": "link"
    "variables": {
            "sourceFolder": "#Dataset0SourceFolder",
            "files": "#Dataset0SelectedFilesPath",
            "file": "@files[0]"
    },

    "url": "https://<host_name>/{{ @instrumentCode }}/{{ @year }}/{{ @proposalIds[0] }}",
    "type": "link",
    "variables": {
            "instrumentCode": "#Instruments[0]Field[uniqueName]",
            "date": "#Dataset[0]Field[creationTime]",
            "proposalIds": "#Dataset[0]Field[proposalIds]"
            "year": "#date_format(@date, yyyy)"
    },
```

## Changes:
Changes to: `src/app/shared/modules/configurable-actions/configurable-action.component.ts`

## Tests included

## Documentation

## Summary by Sourcery

New Features:
- Allow link actions to use template variables in their URLs, including indexed array values, which are resolved at runtime before navigation.


nitrosx (changes requested):
Good feature!!!

The changes assumes that we have only one instrument.
I think we should implement the feature assuming that we can have a list of instruments.
Could you please make the appropriate changes

abdellah257:
> changes assumes that we have only one instrument.
> I think we should implement the feature assuming that we can have a list of instruments.
> Could you please make the appropriate changes

Thanks for the review !

I see your point ! however it seems supporting multiple Instruments for now might be a bit too early, since we are still using `UpdateObsoleteDatasetDTO`, this would require migrating the dataset components to use the new `UpdateDatasetDTO` to have new state management support (store/reduce/actions), so a lot of changes for one PR

nitrosx:
@abdellah257 No problem!!!
Can you still make sure that instruments is an array and the keys to extract the values are consistent with the other properties.
We can stick with one instrument and it would be an array of one, but at least is ready for the transition.
Happy to discuss further

src/app/shared/modules/configurable-actions/configurable-action.component.ts
```diff
@@ -95,16 +89,19 @@ function processSelector(
         .reduce((acc, val) => acc + val, 0),
     // eslint-disable-next-line no-useless-escape
     "#DatasetsField\\[(\\w+)\\]": (m) =>
-      jsonObject.datasets.map((i) => i[m[1]]),
+      jsonObject.datasets?.map((i) => i[m[1]]),
+    "#Instrument\\[(\\d+)\\]Field\\[(\\w+)\\]": (m) => {
```
nitrosx: Change ` #Instrument` to `#Instruments`

src/app/shared/modules/configurable-actions/configurable-action.component.ts
```diff
@@ -351,6 +348,49 @@ export class ConfigurableActionComponent implements OnInit, OnChanges {
...
+        this.actionConfig.url
+          .replace(
+            // Handle {{ #Year( @date ) }},
+            // eslint-disable-next-line no-useless-escape
+            /\{\{\s*\#Year\(([@#]\w+)\)\s*\}\}/g,
+            (_, variableName) => {
+              const date = this.get_value_from_definition(variableName);
+              return String(new Date(date).getUTCFullYear());
+            },
+          )
```
nitrosx: Please see my comment about the configuration.

nitrosx (changes requested):
I was wondering if it would make more sense to change the configuration to the following:
```
"url": "https://<host_name>/{{ @instrumentCode }}/{{ @year) }}/{{ @proposalIds[0] }}",
    "type": "link",
    "variables": {
            "instrumentCode": "#InstrumentField[uniqueName]",
            "date": "#Dataset[0]Field[creationTime]",
            "year"  "#date_format(@date,"+%Y")",
            "proposalIds": "#Dataset[0]Field[proposalIds]"
    },
```
This way the computation is moved to the variable definition and not when you build the url, which requires only substitution.

abdellah257:
- The recursive computation at `actionItems` variables has been addressed, and dependencies between variables solved, so now we can maintain computations at variables field.
- elements selection ei: `@var[0]` can be done either at url level or at variables (see PR description examples)

src/app/shared/modules/configurable-actions/configurable-action.component.ts
```diff
@@ -216,6 +125,156 @@ export class ConfigurableActionComponent implements OnInit, OnChanges {
...
+      if (typeof value === "string") {
+        const matches = value.matchAll(/@(\w+)/g);
+        for (const match of matches) {
+          deps.push(match[1]);
+        }
+      }
+      graph[key] = new Set(deps);
+    });
+    return graph;
+  }
```
nitrosx, line 248: Could you please provide more info why we need this function?

nitrosx (approved):
Next time can you make sure that similar functions can be compared easily, see `processSelector`.

src/app/shared/modules/configurable-actions/configurable-action.component.ts
abdellah257: To resolve action variables we need to resolve their dependencies first, for example:
```json
{
    "a" : "#Dataset[0]Field[instrumentIds]",
    "b" : "@a[0]"
}
```
On this example `b` depends on `a`, so we decide at the `update_status` function to compute `a` first then substitute its value, in this case an Array (`instrumentIds`).
We will then apply the filter on this value.

The functions were added to separate concerns and maintain `processSelector` for static single level resolving.

abdellah257: @nitrosx Documentation about this feature of variables resolving has been added to [`configurable-actions.documentation.md](https://github.com/SciCatProject/frontend/pull/2321/changes#diff-a8d4d6a471c11a67f409958233e429b161043b31ed824a94b39f0f8efae3a008)

PR #2335, opened 2026-04-15
fix: remove vscode settings

emigun:
## Description
I noticed that SciCat frontend overrides my color settings. I think we should be able to set our own colors :)

## Motivation
I like my colors.

## Summary by Sourcery

Chores:
- Delete the committed .vscode/settings.json file so editors no longer enforce shared workspace preferences.


emigun:
Should we remove the whole .vscode folder?

fpotier (approved):
Thanks, for some reason I thought the angular extension was changing the colors and never bothered looking into it.
Imo the whole `.vscode` directory could indeed be removed

PR #2368, opened 2026-05-08
fix(shared-table): track sort/paginator merge subscription for cleanup

alubbock:
## Description
Fix memory leak in `SharedTableComponent` where the `merge(sort.sortChange, paginator.page)` subscription created in `ngAfterViewInit` was not added to the tracked subscriptions array, so it was never cleaned up on component destruction.

## Motivation
The subscription was created with `.subscribe()` but the returned `Subscription` object was discarded rather than pushed to `this.subscriptions`. On every navigation away from and back to a list view, another copy of the subscription was created and kept alive indefinitely. This complements the fix in #2367, which corrected `ngOnDestroy` actually invoking `unsubscribe()`.

## Fixes:

* Wrap `merge(...).subscribe()` in `this.subscriptions.push(...)` so it is cleaned up by `ngOnDestroy`

## Changes:

* `shared-table.component.ts` — subscription is now tracked
* `shared-table.component.spec.ts` — regression test verifies `ngAfterViewInit` registers at least two subscriptions (sort-reset + merge)

## Tests included

## Summary by Sourcery

Track the merged sort/paginator subscription in SharedTableComponent so it is cleaned up on destroy and covered by tests.

Bug Fixes:
- Ensure the sort/paginator merge subscription in SharedTableComponent is tracked for proper cleanup to prevent memory leaks.

Tests:
- Add a regression test confirming ngAfterViewInit registers the sort/paginator merge subscription in the tracked subscriptions list.

### DIFF
--- a/src/app/shared/modules/shared-table/shared-table.component.spec.ts
+++ b/src/app/shared/modules/shared-table/shared-table.component.spec.ts
@@ -203,13 +203,15 @@ describe("SharedTableComponent", () => {
 
   describe("#ngAfterViewInit()", () => {
     it("should add the sort/paginator merge subscription to the tracked list", () => {
-      (component as any).subscriptions = [];
+      const pushSpy = spyOn(
+        (component as any).subscriptions,
+        "push",
+      ).and.callThrough();
+
       component.ngAfterViewInit();
 
-      // Expects at least: sort-reset subscription + merge subscription
-      expect((component as any).subscriptions.length).toBeGreaterThanOrEqualTo(
-        2,
-      );
+      // sort-reset subscription + merge subscription
+      expect(pushSpy).toHaveBeenCalledTimes(2);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


alubbock:
Both issues addressed in the latest commit:

- **Matcher name**: `toBeGreaterThanOrEqualTo` → replaced entirely (see below)
- **Brittleness**: switched from a `length >= 2` check to `spyOn(subscriptions, 'push').and.callThrough()` with `toHaveBeenCalledTimes(2)`, which directly asserts that exactly two subscriptions are registered during `ngAfterViewInit` (sort-reset + merge) without relying on the array's prior state or a magic count.

nitrosx (approved):
@alubbock could you please check the comment from sourceryai and address it.
Other than that, it looks good to me.

alubbock:
@nitrosx comment addressed, sourcery-ai seems happy. Since you've approved, I'll go ahead and merge. Thanks!

PR #2369, opened 2026-05-08
fix(effects): use switchMap for list-fetch effects to prevent stale data

alubbock:
## Description
Replace `mergeMap` with `switchMap` in the four list-fetching NgRx effects that handle user-driven queries (filter changes, sort, pagination).

## Motivation
With `mergeMap`, rapidly changing filters or sort order leaves multiple concurrent HTTP requests in flight simultaneously. Responses can arrive out of order and overwrite newer store state with stale results. For example, a slow first request completing after a faster second one would silently replace the correct data with stale results.

`switchMap` cancels the previous in-flight request when a new action arrives, so only the response to the latest query is ever committed to the store. Angular's `HttpClient` supports cancellation, so the superseded request is also dropped on the network side.

## Fixes:

* `fetchDatasets$` in `datasets.effects.ts`
* `fetchFacetCounts$` in `datasets.effects.ts`
* `fetchMetadataKeys$` in `datasets.effects.ts`
* `fetchProposalDatasets$` in `proposals.effects.ts`

The inner `mergeMap` calls that dispatch multiple actions from a single HTTP response are intentionally unchanged.

## Changes:

* Four one-word operator changes across `datasets.effects.ts` and `proposals.effects.ts`
* Four marble-based regression tests added (one per effect) using the `"-a-b"` / `"--a|"` pattern: two rapid actions with a delayed response, asserting only the second response reaches the store

## Tests included

## Summary by Sourcery

Use cancellable NgRx effects for list-fetching actions to avoid stale data when queries change rapidly.

Bug Fixes:
- Prevent outdated dataset, facet count, metadata key, and proposal dataset responses from overwriting newer store state by switching list-fetch effects to use cancellable operators.

Tests:
- Add marble tests verifying that a second rapid fetch action cancels the in-flight request and only the latest response updates the store for each affected effect.


Junjiequan (approved):

nitrosx:
@alubbock whould you mind resolving the issues with the tests and merge latest master? Thank you so much

PR #2393, opened 2026-05-26
fix: format-number pipe should format booleans

emigun:
## Description
Currently booleans are rendered as empty strings. This PR changes that so that booleans are converted to strings ("true" or "false")

## Motivation
It should be possible to see boolean values

## Tests included

## Summary by Sourcery

Handle boolean values in the formatNumber pipe and extend type guards to treat them as formattable scalars.

Bug Fixes:
- Ensure boolean values are rendered as their string representations instead of empty strings in the formatNumber pipe.

Tests:
- Add and update unit tests to verify boolean handling in both scalar and array inputs for the formatNumber pipe.


minottic (approved):
LGTM, thanks! From a naming perspective the `isFormattableScalar` name is a bit misleading as a boolean is not a formattableScalar (neither a string, at least not in all cases)

nitrosx (approved):

Issue #2386, opened 2026-05-20
Bug: Double click on Datasets header path sends to previous url


abdellah257:
## Summary
 
- If double clicked on the Dataset header path, it will send you back to previous url (even outside scicat)

<img width="151" height="92" alt="Image" src="https://github.com/user-attachments/assets/e02a0c30-402c-4388-a033-951b50e2e135" />

## Expected Behaviour
Send back to `datasets` table page.

## Example:
- You can try it using this public ESS example: [ESS dataset](https://scicat.ess.eu/datasets/20.500.12269%2Fde8f6b13-986f-4f8f-a9b4-cfee16e7012c) and then double click on `Datasets`

Junjiequan:
fixed in https://github.com/SciCatProject/frontend/pull/2398

abdellah257:
Thanks @Junjiequan for the fix !

PR #2398, opened 2026-06-01
fix(breadcrumb): fall back to router navigation when mode is empty

Junjiequan:
## Description
`location.back()` relies on browser history, which causes unexpected redirects (e.g. to google.com) when the page is opened in a new tab or accessed directly via URL.

- **Before:** clicking a breadcrumb in a new tab could redirect to an external URL (browser history fallback)
- **After:** always navigates to the correct fallback route regardless of browser history state

## Fixes:
- Add a guard to check if `mode` is empty before applying it to filters
- If `mode` is empty, navigate directly to the fallback URL without appending mode to filters

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Bug Fixes:
- Prevent incorrect breadcrumb redirects by skipping history.back navigation when the computed view mode is empty and routing directly to the fallback URL.

### DIFF
--- a/src/app/shared/modules/breadcrumb/breadcrumb.component.spec.ts
+++ b/src/app/shared/modules/breadcrumb/breadcrumb.component.spec.ts
@@ -6,11 +6,29 @@ import {
 import { MockStore } from "shared/MockStubs";
 import { BreadcrumbComponent } from "./breadcrumb.component";
 import { Store } from "@ngrx/store";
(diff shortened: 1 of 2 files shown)


Junjiequan:
Fix for https://github.com/SciCatProject/frontend/issues/2386

nitrosx (approved):

PR #2399, opened 2026-06-01
fix: add loading spinner for dynamic table paginator

Junjiequan:
## Description
Added a loading spinner to the dynamic paginator that appears when the facet request is in a pending state.
Currently, this feature is applied only to the datasets and proposals components, as they are the only two components that require facets at the moment.

The example clip uses a 2-second delay on the facets response to mimic a real-world scenario where the facets response is slow.

https://github.com/user-attachments/assets/1fe917d5-dcf9-4ff7-a227-d1a0fa7be9cc

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Add loading-state support to dynamic tables and facet-count fetches so paginators show a spinner while counts are being loaded.

New Features:
- Expose an isLoading flag in dynamic table pagination and render a spinner in the paginator range label when data is loading.

Bug Fixes:
- Track facet count loading state for datasets and proposals and wire it into dataset and proposal tables so pagination reflects ongoing facet-count fetches.

Enhancements:
- Refactor dataset and proposal table subscriptions to combine loading, count, user, and column-setting streams for more consistent table initialization behavior.

### DIFF
--- a/src/app/proposals/proposal-table/proposal-table.component.ts
+++ b/src/app/proposals/proposal-table/proposal-table.component.ts
@@ -27,10 +27,10 @@ import { Store } from "@ngrx/store";
 import {
   selectProposalsWithCountAndTableSettings,
   selectDefaultProposalColumns,
+  selectProposalsFacetCountsIsLoading,
 } from "state-management/selectors/proposals.selectors";
 import { ActivatedRoute, Router } from "@angular/router";
 import {
-  Instrument,
   OutputDatasetObsoleteDto,
   ProposalClass,
 } from "@scicatproject/scicat-sdk-ts-angular";
@@ -54,6 +54,9 @@ export class ProposalTableComponent implements OnInit, OnDestroy {
     selectProposalsWithCountAndTableSettings,
   );
   defaultStoreColumns$ = this.store.select(selectDefaultProposalColumns);
+  isFacetCountsLoading$ = this.store.select(
+    selectProposalsFacetCountsIsLoading,
+  );
   appConfig = this.appConfigService.getConfig();
 
   tableDefaultSettingsConfig: ITableSetting = {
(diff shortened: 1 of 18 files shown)


emigun:
I think it is good that the loading spinner is not activated for each fullfacet request. For example, the sortByColumnAction triggers a request to fullFacet (why...?).

When switching between "Public data" and "My data" setPublicViewModeAction and fetchFacetCountsAction is triggered, causing two fullfacets?

Also switching back and forth between these will cause some problems. If some earlier fullfacet is still running, when it completes I think it will stop the spinner even if the correct fullfacet is not done yet.

See end of video here (spinner disappears, after some time the numbers update):
[Screencast from 2026-06-03 15-21-50.webm](https://github.com/user-attachments/assets/de4b7607-2c64-41e1-a234-5ff6656e599c)

Junjiequan:
@emigun I think the spinner should be triggered by each fullFacets request (isPending=isLoading), and disappears on response, to my knowledge this is the standard behavior? 

The clip you showed looks like a racing condition, which I think is a bug not related to spinner implementation itself. What I would expect is that when second fullFacets request is fired the first fullFacets request should be canceled even it is pending. And it seems fullQuery has the same problem as well from the clip

> `sortByColumnAction triggers a request to fullFacet `

This sounds like a bug as well

emigun:
> @emigun I think the spinner should be triggered by each fullFacets request (isPending=isLoading), and disappears on response, to my knowledge this is the standard behavior?

I didn't check every flow closely, but I noticed that fullfacet start from these actions:
  fetchFacetCountsAction
  setPublicViewModeAction
  sortByColumnAction

but the spinner is only triggered by fetchFacetCountsAction.

Now I'm thinking it might just be a bug that fullfacet is triggered by the other actions? sortByColumnAction should not trigger it (facets don't change) and setPublicViewModeAction triggering it results in duplicate requests.

I agree these problems are not actually related to the spinner.

src/app/proposals/proposal-table/proposal-table.component.ts
```diff
@@ -117,36 +120,47 @@ export class ProposalTableComponent implements OnInit, OnDestroy {
...
-              this.tableName,
-              this.tableDefaultSettingsConfig,
-              userTableConfigColumns,
-              tableSort,
+        .pipe(
+          filter(({ hasFetchedSettings }) => hasFetchedSettings),
+          combineLatestWith(this.isFacetCountsLoading$),
+        )
+        .subscribe(
+          async ([
```
emigun, line 128: Since there is no await this `async` can be removed.

emigun (approved):
Looks good to me (just one small async commnt). We can fix the duplication of requests in a separate PR and change to swithMap instead of mergeMap when calling fullFacet.

PR #2430, opened 2026-06-17
fix: reset search text in add condition dialog and persist search text in globalTextSearch on refresh and navigating between pages

abdimo101:
## Description
This PR fixes two search issues. 

First the search parameters dialog(add condition dialog) now clears the search text when cancelled so the metadata keys list displays properly on reopen. 

Second the dataset table now saves search text to the URL queryParams when applied which matches the behavior in e.g. proposals, files etc, and restores the value visually when navigating back to the datasets page.

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Persist dataset table global text search in URL query parameters and reset metadata key search when cancelling the add-condition dialog.

Bug Fixes:
- Restore global text search value in the datasets table when navigating back via URL query parameters or store state.
- Ensure the metadata keys list in the search parameters dialog is reset by clearing the search term on dialog cancel.

Enhancements:
- Synchronize dataset table text search between URL query parameters and store for consistent behavior across navigations.

### DIFF
--- a/src/app/datasets/dataset-table/dataset-table.component.ts
+++ b/src/app/datasets/dataset-table/dataset-table.component.ts
@@ -31,6 +31,7 @@ import {
   selectTotalSets,
   selectDatasetsInBatch,
   selectDatasetsFacetCountsIsLoading,
+  selectTextFilter,
 } from "state-management/selectors/datasets.selectors";
 import { AppConfigService } from "app-config.service";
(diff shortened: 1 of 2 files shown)


Junjiequan (approved):
lgtm

PR #2478, opened 2026-07-24
fix: url resolving for link configurable action

abdellah257:
## Description
fixes URL resolving for `link` type actions, omitted in a previous commit.

## Changes:
- `link` type configurable actions.

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Bug Fixes:
- Resolve interpolated URLs when opening configurable actions of type `link`.

### DIFF
--- a/src/app/shared/modules/configurable-actions/configurable-action.component.ts
+++ b/src/app/shared/modules/configurable-actions/configurable-action.component.ts
@@ -444,7 +444,10 @@ export class ConfigurableActionComponent
   }
 
   private typeLink() {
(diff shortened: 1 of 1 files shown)


Junjiequan (approved):
I guess the original intention was to provide pure url without any variables in the URL? but I don't see any downside processing url via interpolate function, as the value conversion is restricted to curly brackets only.

One concern I have is the url which is not encoded, and might cause errors in some cases, but it's not related to this PR. 

```js
  private typeXhr() {
    const url = this.interpolate(this.actionConfig.url);

    fetch(url, {
      method: this.actionConfig.method || "POST",
    })
```

PR #2507, opened 2026-08-21
refactor(services): add shared copy-to-clipboard service

rkweehinzmann:
## Description
a new copy service is introduced as shared service and used in published data details. 

## Motivation
More convenience: Goal was to make the doi copyable just like the pids in dataset details. 

## Changes:
introduce new service at central place (app/shared/service) to use in published data for doi field. 

* changes made:
* added files as was suggested by AI: introduce a new service in shared/service, also in index.ts to export the function then where it is used in publisheddata-details.components.ts and its html.

## Other
For maintainability reasons one can follow up replacements of other places (see screenshot) where this functionality was used. Main bottleneck is testing that functionality remains in all cases the same.

Screenshot from github bot
Screenshot 2026-08-21 at 09 02 07

## Tests included

## Backend version
- which version of the backend is required: not needed

## Summary by Sourcery

Centralize clipboard functionality and make published data DOIs copyable from their details view.

New Features:
- Add a copy-to-clipboard action for DOI values in published data details.

Enhancements:
- Centralize clipboard copying and user feedback in a shared service.
- Update existing dataset, job, and user-settings copy actions to use the shared clipboard service with support for unavailable or failed clipboard operations.

### DIFF
--- a/src/app/datasets/dataset-detail/dataset-detail/dataset-detail.component.ts
+++ b/src/app/datasets/dataset-detail/dataset-detail/dataset-detail.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, OnDestroy, Inject } from "@angular/core";
+import { Component, OnInit, OnDestroy } from "@angular/core";
 import { ENTER, COMMA, SPACE } from "@angular/cdk/keycodes";
 import { MatChipInputEvent } from "@angular/material/chips";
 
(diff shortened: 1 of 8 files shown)


rkweehinzmann:
code has been added to use the new ClipboardService in other places too: pid, scicat token and jobId. Also the tests were adopted.

belfhi:
@rkweehinzmann it should be `refactor: add copy...` not `feat(refactor)` 😉

PR #2515, opened 2026-08-26
fix: return generated JWT token for configurable actions

abdellah257:
## Description
Replace the JWT sdk token, with a freshly generated JWT that contains information from the User Profile.

## Motivation
The `#token` already returns the jwt sdk token, the generated `#jwt` is supposed to include profile information (e.g AccessGroups)
```ts
private authorizationTokens = {
    "#jwt": () => this.jwt,
    "#token": () => this.authService.getToken()?.id,
    "#tokenSimple": () => this.authService.getToken()?.id,
    "#tokenBearer": () => `Bearer ${this.authService.getToken()?.id}`,
    "#uuid": () => uuidv4(),
  };
```

## Tests included

## Documentation

## Backend version

## Summary by Sourcery

Bug Fixes:
- Return the freshly generated user JWT for configurable action `#jwt` substitutions instead of the SDK access-token JWT.

### DIFF
--- a/src/app/shared/modules/configurable-actions/configurable-action.component.ts
+++ b/src/app/shared/modules/configurable-actions/configurable-action.component.ts
@@ -12,6 +12,7 @@ import { DatePipe } from "@angular/common";
 import {
   Configuration as ApiConfiguration,
   DatasetClass,
+  UsersService,
 } from "@scicatproject/scicat-sdk-ts-angular";
 import {
   ActionButtonStyle,
@@ -81,6 +82,7 @@ export class ConfigurableActionComponent
   form: HTMLFormElement | null = null;
 
   constructor(
+    private usersService: UsersService,
     private authService: AuthService,
     private configService: AppConfigService,
     private store: Store,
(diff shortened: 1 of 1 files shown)


minottic:
thanks for the change @abdellah257 . I had actually refactored this to use the token stored as doing a fetch triggers an http command which invokes a snackbar, so whenever submitting an action you get one extra snackbar "success" that is a bit counterintuitive for the user. I would lean towards passing the user groups in the action context from the parent component if you need the groups, but happy to change my mind1

Junjiequan (approved):
This is the exact same change I was about to do, thanks

Junjiequan:
@minottic Can we merge this one first and then refactor it after, the previous change is breaking our external services which relys on the groups

minottic:
ofc, sorry for that!

minottic (approved):

abdellah257:
Thanks for the feedback. @minottic to avoid your issue, I was thinking maybe we can add this Jwt token to the user state since it seems to be used on other places too (`datafiles`). we can't submit the groups in actions separately from the token for security reasons.

